PR #268, opened 2018-12-05
STScI doc theme

robelgeda:
This PR introduces the STScI doc theme. It also fixes some minor doc-string errors. 
screen shot 2018-12-05 at 5 53 56 am

PR #269, opened 2018-12-06
Dummy PR: Travis

robelgeda:
This PR is for testing Travis CI behavior.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,8 @@ env:
         - NUMPY_VERSION=stable
         - ASTROPY_VERSION=stable
         - SETUP_CMD='test'
-        - PIP_DEPENDENCIES='jwxml nbsphinx Sphinx==1.5.6'
-        - CONDA_DEPENDENCIES='scipy matplotlib six cython'
+        - PIP_DEPENDENCIES=''
+        - CONDA_DEPENDENCIES='scipy matplotlib six'
 
         # For headless testing of code that imports PyPlot:
         - MPLBACKEND=Agg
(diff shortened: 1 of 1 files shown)


robelgeda:
That fixed the issue, now remove each part until Travis breaks...

robelgeda:
At this point it looks like the issue is due to MatPlotLib

robelgeda:
Because the MPLBACKEND sets the backend for matplotlib. I have actually run into this when working on the Viz tools. Its caused by requirements for matplotlib not being met. Speaking of which....
https://matplotlib.org/2.0.2/users/installing.html#required-dependencies
screen shot 2018-12-05 at 11 54 50 pm

screen shot 2018-12-05 at 11 51 46 pm

That solves the mystery for why this fixes the bug... but as to why Travis didn't tell us this ¯\_(ツ)_/¯

mperrin:
I’m not sure I follow the point of what you’re highlighting in the package versions - matplotlib 3 requires numpy >1.10, and matplotlib 2 requires numpy >1.7, both of which are easily met by our installing numpy 1.15. So that seems fine? Are you suggesting there’s some version incompatiblity? Sorry I’m not following along...

That said, what’s the minimum versions of all this that the JWST data analysis tools are targeting? Do we have any sort of standard policy for how far back we support? I’d be happy to drop support for some of these older versions. Now would actually be a good time to do that - this is the first Python-3-only release of Poppy and WebbPSF, so it’s distuptive anyways.

robelgeda:
I take that back... that requirement is for matplotlib 2.0.2. I am not sure why the backend was set to something wrong to start with... I can see pyqt being installed and I know it uses `AGG`. Here is some info on the available backends:
https://matplotlib.org/tutorials/introductory/usage.html#backends

robelgeda:
Note that #268 and #267 failed when reran. I merged master now and will check back.

robelgeda:
Looks like they pass once that line of code is added!

PR #270, opened 2018-12-06
Fix travis by setting MPLBACKEND

robelgeda:
I did an extensive test diagnostics via #269 (I reset the branch multiple times so not all the commits show). I finally used the WebbPSF's `.travis.yml` file to see if it worked. I then removed each change one by one to find that the `MPLBACKEND` definition fixes the issue we are seeing. This means that the test environment did not know which MPL backend to use. This was weird because Travis showed no errors when it crashed; but now we realize it segfaulted when trying to use/compile MPL & PyPlot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,9 @@ env:
         - SETUP_CMD='test'
         - PIP_DEPENDENCIES=''
         - CONDA_DEPENDENCIES='scipy matplotlib six'
+
+        # For headless testing of code that imports PyPlot:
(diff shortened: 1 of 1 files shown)


mperrin:
What’s weird is, the versions on travis with older numpy versions were all working ok. I guess because they were also getting older matplotlib versions too, which wasn’t subject to this issue? I don’t understand entirely why it was failing like that, without showing errors like you say. But hey, a fix is a fix in any case. Thanks for sleuthing this out!

Issue #272, opened 2018-12-06
add CircularPhaseMask class


mperrin:
Needed for Zernike WFS simulator for optics lab.

PR #273, opened 2018-12-07
Add CircularPhaseMask coronagraph

mperrin:
Adds the new feature requested in #272, a CircularPhaseMask class useful for either old-school Roddier & Roddier coronagraphs, or a Zernike wavefront sensor.


mperrin:
@ivalaginja and @sdwill  - just FYI here's an example of adding a new class to poppy to implement a particular coronagraphic element. As discussed yesterday in the lab sim meeting. 

This core idea in this class is about as simple as you would expect; much of the code is nice-to-have but nonessential extras for display, giving better warnings on certain error conditions, etc.

PR #276, opened 2018-12-08
better default parameters for OpticalSystem.input_wavefront 

### LINKED ISSUE #246: OpticalSystem.input_wavefront default sampling should look to OpticalSystem attributes first
<a href="https://github.com/mperrin"></img></a> **Issue by [mperrin](https://github.com/mperrin)**
_Friday Apr 06, 2018 at 18:32 GMT_
_Originally opened as https://github.com/mperrin/poppy/issues/246_

----

Currently, the logic in `input_wavefront` for determining the sampling of the wavefront array is to first look at the pupil size and number of pixels for the first plane in the optical system, and only if those are None does it look to the attributes of the OpticalSystem instance itself. This violates the principle of least astonishment. If the user has specifically set `npix` or `pupil_diam` on the OpticalSystem, that should be the first place that is checked when creating an wavefront, and should override whatever the first OpticalElement has.  

As a nice side benefit, this also will make the Fraunhofer interface more consistent with the Fresnel one.

mperrin:
This PR fixes #246:
>Currently, the logic in input_wavefront for determining the sampling of the wavefront array is to first look at the pupil size and number of pixels for the first plane in the optical system, and only if those are None does it look to the attributes of the OpticalSystem instance itself. This violates the principle of least astonishment. If the user has specifically set npix or pupil_diam on the OpticalSystem, that should be the first place that is checked when creating an wavefront, and should override whatever the first OpticalElement has.

The result is that input wavefront sizes are determined as following. 
1. If `npix` or `pupil_diameter` are defined on the OpticalSystem as a whole, those values take precedence.
2. Otherwise check if the first OpticalElement in the system defines the sampling number of pixels (by having a `.shape`) and/or has a `.pupil_diam` attribute. 
3. If none of the above, default to `npix=1024` for the pixel grid, or raise an error if nothing defines the size of array to sample.

As noted in #246, this also has the benefit of making OpticalSystem consistent with FresnelOpticalSystem in terms of how the wavefront sampling is defined. 

Relatedly but distinct, I also went ahead and updated the default wavelength to 1 micron everywhere, rather than a mix of 1 and 2 microns.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -1116,7 +1116,7 @@ class FresnelOpticalSystem(OpticalSystem):
         return inwave
 
     @utils.quantity_input(wavelength=u.meter)
-    def propagate_mono(self, wavelength=2e-6 * u.meter,
+    def propagate_mono(self, wavelength=1e-6 * u.meter,
                            normalize='first',
                            retain_intermediates=False,
                            retain_final=False,
--- a/poppy/optics.py
+++ b/poppy/optics.py
@@ -215,7 +215,7 @@ class AnalyticOpticalElement(OpticalElement):
             return output_array
 
     @utils.quantity_input(wavelength=u.meter)
-    def display(self, nrows=1, row=1, wavelength=2e-6 * u.meter, npix=512, grid_size=None,
+    def display(self, nrows=1, row=1, wavelength=1e-6 * u.meter, npix=512, grid_size=None,
                 what='intensity', **kwargs):
         """Display an Analytic optic by first computing it onto a grid...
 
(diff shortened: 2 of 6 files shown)


mperrin:
I am confused about the test coverage reports here. Coveralls above is reporting that the test coverage drops by 1%, which I can't understand why that's happening. The lines which lose coverage are all in various branch paths of instantiating FITSOpticalElement (see [this page on coveralls](https://coveralls.io/builds/20552472/source?filename=poppy%2Fpoppy_core.py)) but that doesn't seem related at all to what's in this PR. Weird.

mperrin:
Added some additional test cases (for stuff that was to be honest not particularly related to this PR). Rebased on current master after the most recent other PRs were merged. Force pushed the rebased PR here. Let's see how Coveralls and Travis think of this now.

PR #278, opened 2018-12-10
doc theme patch

robelgeda:
Remove outdated theme settings.


mperrin:
Looks good. And I added you as an admin on Readthedocs. I'm going to merge this now and we can see if it triggers the doc build there correctly.

PR #281, opened 2018-12-12
Update URL for spacetelescope repo

shanosborne:
There was just 1 URL that had to be updated in the poppy docs for spacetelescope - the rest had already been updated


mperrin:
The above coveralls report makes no sense - if zero lines changed coverage, how does the overall coverage decrease 0.6%?!? Not sure this tool is working as intended...

shanosborne:
Oh weird... I have no idea. That'd be a good thing to look into!

mperrin:
I re-ran the coveralls and now it shows a 1.4% increase instead of -0.6% decrease, for the exact same PR. So that’s... not super helpful. I’m going to ignore this for now and just merge this. But post-holidays and AAS maybe we can find someone here who understands Coveralls better than we do...

PR #286, opened 2018-12-21
Reassign BUNIT header keyword immediately after rescaling units to m

### LINKED ISSUE #285: Re-assign BUNIT hdr keyword sooner
When an OPD file is loaded into a `FITSOpticalElement`, it is quite immediately re-scaled to be in units of meters (`poppy_core.py`, lines 2381-2387). However, the `BUNIT` header keyword corresponding to the OPD is not re-written to reflect this. This confused me, as I was looking at OPD arrays and seeing perturbations on the scale of meters, while the `opd_header` was still telling me the units were in microns.

The `BUNIT` keyword is eventually re-written if the OPD is written to an HDUList object in `webbpsf.opds.as_fits()`, but this isn't helpful for someone looking at the OPD array and header objects.

I think it would be wise to re-write this keyword to reflect that the OPD is now in units of meters as soon as the OPD is re-scaled. If there are no disagreements, I would be happy to submit a quick PR for it.

laurenmarietta:
Closes #285.

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -2375,6 +2375,7 @@ class FITSOpticalElement(OpticalElement):
                 self.opd *= 1e-6
             elif opdunits in ('nanometer', 'nm'):
                 self.opd *= 1e-9
+            self.opd_header['BUNIT'] = 'meter'
 
             if len(self.opd.shape) != 2 or self.opd.shape[0] != self.opd.shape[1]:
                 _log.debug('OPD shape: ' + str(self.opd.shape))
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -2375,7 +2375,8 @@ class FITSOpticalElement(OpticalElement):
                 self.opd *= 1e-6
             elif opdunits in ('nanometer', 'nm'):
                 self.opd *= 1e-9
-            self.opd_header['BUNIT'] = 'meter'
+            if self.opd_header is not None:
+                self.opd_header['BUNIT'] = 'meter'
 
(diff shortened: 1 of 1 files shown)


mperrin:
Hi @laurenmarietta I happened to notice this PR was still open.  I couldn't remember why I hadn't merged it, but then noticed the CI test is failing. Tried re-running that on Travis again but it still failed.  

This is totally non-urgent (clearly this has been sitting on the back burner for 2 months now with neither of us remembering about it) but could you please take a look sometime at why the test is failing? No rush at all on this; I just don't want to leave this PR hanging open too long. Thanks

laurenmarietta:
Thanks for the ping on this - I needed to add an if statement around the header keyword assignment!

There is still one failing test, but at first glance I don't think it has to do with the 2 lines of code I've added:
```
ImportError while loading conftest '/tmp/poppy-test-jugy57_p/lib/python3.6/site-packages/poppy/conftest.py'.
poppy/conftest.py:5: in <module>
    from astropy.tests.pytest_plugins import *
E   ModuleNotFoundError: No module named 'astropy.tests.pytest_plugins'
```

If you disagree though, I can try to investigate.

mperrin:
I agree, I saw that error also on an unrelated PR, and it's clearly due to some astropy rearrangements splitting their plugins off into another package. @shanosborne could you take a look at whatever we need to do to update webbpsf and poppy to work with the latest astropy?

mperrin:
The failing Travis case is some unrelated astropy thing, I agree, so I'm going to go ahead and merge this now.

Issue #287, opened 2019-02-07
Notebook examples use deprecated function names


corcoted:
Some of the example notebooks no longer run since the name changes in version 0.8.0 because they use deprecated function names (e.g. calcPSF).

Should just need to rename.

corcoted:
Submitted pull request #288

corcoted:
It also looks like some of the tests still use the old function names, but I haven't been able to run them to check (config problem on my end).

mperrin:
I found and fixed some instances of `_getOpticalSystem` in one of the test files, but that's all I can find in the tests. Are there other instances of any old function names that I'm missing?

mperrin:
Fixed by #288

PR #288, opened 2019-02-07
Replaced function names in docs and example notebooks to remove deprecated names

corcoted:
I ran all of the example jupyter notebooks with poppy version 0.8.0 and replaced any function names that failed.  I also found the old function names in some docs and replaced those.


corcoted:
To fix issue #287

mperrin:
Thanks very much @corcoted. 
On Travis, one test case is failing but it's an unrelated issue due to an API change in astropy upstream, not anything related to this. So I'm going to go ahead and merge this. Much appreciated.

Issue #280, opened 2018-12-10
Passing wavefronts from one OpticalSystem to another


mperrin:
This one's tricky. Issue raised by @sdwill:

Right now there is no way to "connect together" multiple optical systems, in the sense of taking an output wavefront from a propagation through one system, and feeding it into a propagation through another system. This could be useful in a variety of contexts including
 - Building separate, modular optical system models for e.g. telescope and instrument separately, then sticking them together
 - Combining different kinds of propagators, such as a mixed Fraunhofer and Fresnel calculations
 - Running a calculation for a situation in which the starting point is not a simple plane wave as assumed by the existing `input_wavefront` function. 

The complex wavefront output from a given optical system can be obtained using the `retain_final` parameter to `calc_psf`. But there is no easy way to hand that in to another system. 

Possible enhancements to address this might include one or more of the following alternatives: 
 - Example code for how to replace `input_wavefront` with a function that grabs some other existing wavefront as a starting point. Potentially such a function might even run a completely different other system recursively to get its starting wavefront.
 - A container optical system class, notionally `CompoundOpticalSystem`, which could work similarly to `CompoundOpticalElement` to iterate over multiple other systems and act to glue them together into something that acts like a larger whole. 
 - An ability to directly "add together" OpticalSystem instances, and returns a new optical system instance containing the concatenation of the two sets of optics  
 - A new "hybrid" optical system class which allows mixed Fraunhofer & Fresnel propagations? 

Any of the above might have to deal with potential issues of resampling, padding, and other possible inconsistencies of pixel grid sampling between the different stages of calculation.

mperrin:
Thought about this some more. Possibly the right mental model here is to think of optical systems as linear operators acting on wavefronts.  That's true mathematically, but not in how the code here is currently implemented. One could imagine a function on an optical system, maybe call it `propagate_through`, which takes a wavefront and returns another wavefront: 
```
wave_out = osys.propagate_through(wave_in)
```
That would make it easy to compose optical systems arbitrarily: 
```
wave_out = osys2.propagate_through(osys1.propagate_through(wave_in))
```
The current function `propagate_mono` comes closest to that, but doesn't have exactly the right semantics because it creates a new input wavefront rather than accepting one as a parameter, and it hands back a PSF rather than a wavefront. 

[...]

PR #290, opened 2019-02-11
Refactoring to enable chaining of optical systems, and mixed Fresnel/Fraunhofer

mperrin:
A relatively substantial refactoring of some core propagation functions. See #280 for more on the motivating use case: passing the outputs of one propagation into the start of another propagation. For instance, this could be used to perform a mixed calculation using Fresnel in some parts and Fraunhofer in others. 

This PR implements this via the following. 

- Refactors most of the core OpticalSystem propagation logic from `propagate_mono` into a new function, tentatively called `propagate_through`, which is a linear operator acting on Wavefront objects. The old `propagate_mono` now becomes a wrapper that creates an input wavefront and passes it through that new function, then handles the output formatting. This preserves the old API and functionality for all existing code, while adding the desired new functionality.
- Refactors `Wavefront` and `FresnelWavefront` to both be children of a parent abstract base class, `BaseWavefront`. This is cleaner than before, in which `FresnelWavefront` was a child of `Wavefront` but with different semantics. This also makes it easier to test which kind of wavefront you're dealing with - i.e. before `isinstance(Wavefront)` would be true for both kinds, now it's only true for pure `Wavefront` instances but not `FresnelWavefront` instances.
- Implements  new functions for casting back and forth between those two kinds of wavefronts. 
- Implements a new class `CompoundOpticalSystem` to easily chain multiple optical systems. It's the same kind of container and iterator idea as for `CompoundAnalyticOptic`, but applied to optical systems using the above. 
- Some supporting changes to wavefront display during calculations, to let it work as desired inside a chained calculation (i.e. the display of a later optical system adds more rows of wavefront plots rather than overwriting the first set of plots)

There are also some less-directly-related changes included in this PR: 
- Improves support for Detectors, in particular for resampling a wavefront to match the detector sampling even if the wavefront is already at the right plane so there's not strictly any propagation needed. 
- A variety of minor edits, comment cleanups, and semi-unrelated fixes made in the course of all the above, while trying to get this working for HiCAT multi-DM sims. 

Apologies in advance but this PR is going to be a pain to read. I recommend reading the part in `poppy_core` and ignoring everything else to start. Note, Github seems to want to hide that by default because it's a large diff, so be sure to click to expand it.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -1139,50 +1139,6 @@ class FresnelOpticalSystem(OpticalSystem):
         inwave._display_hint_expected_nplanes = len(self)     # For displaying a multi-step calculation nicely
         return inwave
 
-#    @utils.quantity_input(wavelength=u.meter)
-#    def propagate_mono(self, wavelength=1e-6 * u.meter,
-#                           normalize='first',
-#                           retain_intermediates=False,
-#                           retain_final=False,
-#                           display_intermediates=False):
-#        """Propagate a monochromatic wavefront through the optical system, via Fresnel calculations.
-#        Called from within `calc_psf`.
-#        Returns a tuple with a `fits.HDUList` object and a list of intermediate `Wavefront`s (empty if
-#        `retain_intermediates=False`).
-#
-#        Parameters
-#        ----------
-#        wavelength : float
-#            Wavelength in meters
-#        normalize : string, {'first', 'last'}
-#            how to normalize the wavefront?
-#            * 'first' = set total flux = 1 after the first optic, presumably a pupil
-#            * 'last' = set total flux = 1 after the entire optical system.
-#            * 'first=2' = set total flux = 2 after the first optic (used for debugging only)
-#        display_intermediates : bool
-#            Should intermediate steps in the calculation be displayed on screen? Default: False.
-#        retain_intermediates : bool
-#            Should intermediate steps in the calculation be retained? Default: False.
-#            If True, the second return value of the method will be a list of `poppy.Wavefront` objects
-#            representing intermediate optical planes from the calculation.
-#        retain_final : bool
-#            Should the final complex wavefront be retained? Default: False.
-#            If True, the second return value of the method will be a single element list
-#            (for consistency with retain intermediates) containing a `poppy.Wavefront` object
-#            representing the final optical plane from the calculation.
-#            Overridden by retain_intermediates.
-#        Returns
-#        -------
-#        final_wf : fits.HDUList
-#            The final result of the monochromatic propagation as a FITS HDUList
-#        intermediate_wfs : list
-#            A list of `poppy.Wavefront` objects representing the wavefront at intermediate optical planes.
-#            The 0th item is "before first optical plane", 1st is "after first plane and before second plane", and so on.
-#            (n.b. This will be empty if `retain_intermediates` is False and singular if retain_final is True.)
-#        """
-#        if poppy.conf.enable_speed_tests:
-#            t_start = time.time()
-
     def propagate_through(self,
                           wavefront,
                           normalize='none',
@@ -1250,7 +1206,6 @@ class FresnelOpticalSystem(OpticalSystem):
             return wavefront, intermediate_wfs
         else:
             return wavefront
-
 
     def describe(self):
         """ Print out a string table describing all planes in an optical system"""
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -970,30 +970,8 @@ class FresnelWavefront(BaseWavefront):
             _log.debug("Wavefront is already at desired pixel scale {:.4g}.  No resampling needed.".format(self.pixelscale))
             return
 
-        _log.info("Resampling wavefront to detector with {} pixels and {}. Zoom factor is {:.4g}".format(
-            detector.shape, detector.pixelscale, pixscale_ratio ))
+        super(FresnelWavefront, self)._resample_wavefront_pixelscale(detector)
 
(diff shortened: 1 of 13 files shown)


mperrin:
@douglase, @sdwill I'd appreciate any feedback you have on this one.

poppy/fresnel.py
```diff
@@ -982,32 +965,52 @@ def _resample_wavefront_pixelscale(self, detector):
...
-        new_wf_real = scipy.ndimage.zoom(self.wavefront.real, pixscale_ratio, order=detector.interp_order)
-        new_wf_imag = scipy.ndimage.zoom(self.wavefront.imag, pixscale_ratio, order=detector.interp_order)
-        new_wf = new_wf_real + 1.j*new_wf_imag
+    @classmethod
+    def from_wavefront(cls, wavefront):
+        """Convert a Fraunhofer type wavefront object to a Fresnel one
 
-        _log.debug("Cropping/padding resampled wavefront to detector shape")
-        new_wf = utils.pad_or_crop_to_shape(new_wf, detector.shape)
+		Note, this function implicitly assumes this wavefront is at a
```
douglase: it might be useful to at least have a placeholder for image plane conversion (where the user provides the plate scale).

poppy/poppy_core.py
```diff
@@ -1032,6 +1078,103 @@ def coordinates(self):
...
+        new_wf.wavefront = utils.pad_or_crop_to_shape(wf.wavefront, new_wf.shape)
+        # Copy over misc internal info
+        if hasattr(wf, '_display_hint_expected_nplanes'):
+            new_wf._display_hint_expected_nplanes = wf._display_hint_expected_nplanes
+        new_wf.current_plane_index = wf.current_plane_index
+        new_wf.location = wf.location
+
+        return new_wf
+
+class Wavefront(BaseWavefront):
```
douglase: I think it's good to keep this name, otherwise a lot of legacy code will break

poppy/poppy_core.py
```diff
@@ -78,25 +78,10 @@ def _wrap_propagate_for_multiprocessing(args):
...
-        * `wf.pixelscale`,   a  scale in arcsec/pixel
-
-
-    Use the `wf.propagate_to()` method to transform a wavefront between conjugate planes. This will update those
-    properties as appropriate.
-
-    By default, `Wavefronts` are created in a pupil plane. Set `pixelscale=#` to make an image plane instead.
+class BaseWavefront(object):
+    """ Abstract base wavefront class.
+    Don't use this class directly, use either Wavefront or FresnelWavefront
```
douglase: "it's not recommended to use this class directly"?

poppy/tests/test_fresnel.py
```diff
@@ -533,3 +533,123 @@ def test_detector_in_fresnel_system(npix=256):
...
+        reduce test run time.
+    """
+
+    import poppy
+
+    opt1 = poppy.SquareAperture()
+    opt2 = poppy.CircularAperture(radius=0.55)
+
+    # a single optical system
+    osys = poppy.FresnelOpticalSystem(beam_ratio=0.25, npix=npix)
```
douglase: beam ratio seems like it adds confusion for folks not coming from PROPER, should we start defaulting to oversample and just have an optional keyword for beam ratio, which throws an error if both are defined?

poppy/poppy_core.py
mperrin: Absolutely! There was never a chance I'd break back-compatibility in a major way with this. Keeping compatibility with existing code is a high priority. And all the unit tests enforce this.

poppy/tests/test_fresnel.py
mperrin: I'm fine with that - and I agree having a more consistent API between the two versions would be beneficial. 
Right now the FresnelOpticalSystem class actually doesn't have an oversample keyword - it just has beam_ratio. I can't even remember why we did it that way since FresnelWavefront has oversample.  (In retrospect I also don't even like the keyword name 'oversample' in this sense, since 2 for Nyquist is well sampled rather than oversampled. But that too would break a lot of legacy code to change.)

That said, while I agree with this proposed change, I think it ought to be a separate PR from this one.

mperrin:
>I've tried to do this the hard way, making two different wavefronts and then copying the complex array from one into the other, but you can loose book-keeping parameters easily.

In some sense that's what this code does, just with two pre-created matched conversion functions to hopefully not lose the book-keeping. 

> I think it would be useful to be able to switch from the image plane by allowing the user to input a known plate scale conversion, for example for some systems the Fraunhofer MFT saves a lot of time getting to an image plane that you might want to treat downstream as a coronagraph, but maybe there's a more subtle reason why you would want to stay in the Fraunhofer assumption until the next pupil?

No deep reason. I just was limiting the scope of this first version of the conversion code, to the kind of conversion I need right now for HiCAT. We could certainly later add an image plane conversion as you suggest. 

I'm now leaning towards naming the new function just `propagate` rather than `propagate_through`. It reads better: 

```
system.propagate(wavefront)
```
seems right, while
```
system.propagate_through(wavefront)
```
is backwards. (The wavefront propagates through the system, not the other way around.)

mperrin:
@douglase @sdwill Just wanted to check whether there's any chance you might have time to take a look at this between now and say early next week sometime. If not then that's fine. Just let me know either way, please?  I'm trying to plan a little around other tasks and lab-related modeling I'm doing now, and it would help to know whether I should wait for any further comments/feedback or should just keep proceeding with this.

douglase:
I didn't see any red flags, cool with merging, I will certainly bug you if things break.

sdwill:
Took a close look at the code for CompoundOpticalSystem and it looks fine to me.  I'm all for merging.

PR #294, opened 2019-02-18
Fix astropy.tests.pytest_plugins error

### LINKED ISSUE #289: Upstream astropy removal of astropy.tests.pytest_plugins
Apparently `astropy.tests.pytest_plugins` is moved someplace else in the next astropy. Not sure what change(s) we need to make in response to this. But right now our Travis builds are failing on the astropy development test case: 

See for instance this failing Travis build: 
https://travis-ci.org/spacetelescope/poppy/jobs/490316242

@shanosborne @robelgeda could one of you please take a look at this? @robelgeda have you seen anything similar with any of the other tools packages? Thanks.

shanosborne:
Fixes #289 problem with astropy.tests.pytest_plugins. See links to reasoning behind this fix here: https://github.com/spacetelescope/poppy/issues/289#issuecomment-464885028

### DIFF
--- a/poppy/conftest.py
+++ b/poppy/conftest.py
@@ -2,7 +2,11 @@
 # by importing them here in conftest.py they are discoverable by py.test
 # no matter how it is invoked within the source tree.
 
-from astropy.tests.pytest_plugins import *
+try:
+    from astropy.tests.plugins.display import *
+    from astropy.tests.helper import *
+except ImportError:
+    from astropy.tests.pytest_plugins import *
 
 ## Uncomment the following line to treat all DeprecationWarnings as
 ## exceptions


shanosborne:
@robelgeda - just to keep you in the loop

mperrin:
Thanks @shanosborne. 

The fix looks good - though I'm a little confused by the report from coveralls, which appears to say that this decreases test coverage in a whole bunch of unrelated files? I've seen apparently-buggy behavior like this from coveralls several times lately. Starting to wonder if we should switch over to codecov.io instead, which I've seen used on some of the astropy repos etc.  I find it hard to believe that this fix would cause a 70-line reduction in test coverage in matrixDFT.py -- do you agree, or is there maybe some chance that's a real side effect of this? 

I'm going to go ahead and merge this now, but @shanosborne @robelgeda let me know if either of you have thoughts on more reliable code coverage tools...

PR #296, opened 2019-02-21
Dummy PR: Checking coverage report

shanosborne:
Dummy PR to check/fix coverage decrease from PR #294

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ env:
         - NUMPY_VERSION=stable
         - ASTROPY_VERSION=stable
         - SETUP_CMD='test'
-        - PIP_DEPENDENCIES=''
+        - PIP_DEPENDENCIES='numexpr'
         - CONDA_DEPENDENCIES='scipy matplotlib'
 
         # For headless testing of code that imports PyPlot:


mperrin:
I see you're just testing this - FYI, it's intentional that numexpr isn't in the main pip_dependencies at at the top of the file. We want to have different test cases both with and without it. So it's not installed a the top by default, but is installed in one of the specific test cases. See line 36 of the travis config file

shanosborne:
Yes I did notice that. What prompted that edit was that it seems like most/all of the files that had the coverage drop in the previous PR came from the line `if accel_math._USE_NUMEXPR:` being `False` when in the past it had been `True`. And it looked like part of the reason it would be set to false was if the import of `numexpr` failed. So I just wanted to look into how that import happens in travis. But that's good to know that there was a reason it was only included in certain test cases.

shanosborne:
Also @mperrin do you know why the coverage test isn't run at the same time as travis here? They run at the same time in webbpsf don't they?

mperrin:
> Also @mperrin do you know why the coverage test isn't run at the same time as travis here? They run at the same time in webbpsf don't they?

Why do you say they're not running at the same time? My understanding of how this works is that Travis runs the actual coverage test as one or more of its jobs (see the lines in the travis.yml file with `test --coverage`), the results from those get uploaded to Coveralls, then Coveralls computes the report and posts a comment here onto Github. So it's sort of all part of one series of connected steps.

shanosborne:
> > Also @mperrin do you know why the coverage test isn't run at the same time as travis here? They run at the same time in webbpsf don't they?
> 
> Why do you say they're not running at the same time? My understanding of how this works is that Travis runs the actual coverage test as one or more of its jobs (see the lines in the travis.yml file with `test --coverage`), the results from those get uploaded to Coveralls, then Coveralls computes the report and posts a comment here onto Github. So it's sort of all part of one series of connected steps.

Whoops that was my bad - the coverage report hadn't shown up yet when I posted that comment so I thought it hadn't run, but now I see it. So it's all good.

PR #299, opened 2019-02-26
Update display method in OpticalElement class 

shanosborne:
This PR is a fix for a reported issue in a STARS ticket.

Issue: "In the file `poppy_core.py`, in `class OpticalElement`, the `display` method has an optional positional argument called `title` which defaults to `None`. Presumably this is so the user can pass in the title to be used in the `display`. However, passing in a title does not work, as the code steps on anything that is provided."

I fixed it so now if `title=None`, the default title structure of "{what your plotting} for {`self.name`}" will remain. But if `title` is not `None`, that input is used for the `title` (or `suptitle` in the case of `what=='both'` since you have 2 subplots)

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -2429,17 +2429,19 @@ class OpticalElement(object):
                          colorbar_orientation=colorbar_orientation, title=None, opd_vmax=opd_vmax,
                          nrows=nrows)
             ax2.set_ylabel('')  # suppress redundant label which duplicates the intensity plot's label
+            if title is not None:
+                plt.suptitle(title)
             return ax, ax2
         elif what == 'amplitude':
             plot_array = ampl
-            title = 'Transmissivity'
+            default_title = 'Transmissivity'
             cb_label = 'Fraction'
             cb_values = [0, 0.25, 0.5, 0.75, 1.0]
             cmap = cmap_amp
             norm = norm_amp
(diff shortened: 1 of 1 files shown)


shanosborne:
E.g. `what='opd'`, `title='Test Title'`

screen shot 2019-02-26 at 2 38 26 pm

E.g. `what='both'`, `title='Test Title'`
screen shot 2019-02-26 at 2 38 02 pm

poppy/poppy_core.py
```diff
@@ -2472,7 +2474,10 @@ def display(self, nrows=1, row=1, what='intensity', crosshairs=False, ax=None, c
         utils.imshow_with_mouseover(plot_array, ax=ax, extent=extent, cmap=cmap, norm=norm,
                                     origin='lower')
         if nrows == 1:
-            plt.title(title + " for " + self.name)
+            if title is None:
+                plt.title(default_title + " for " + self.name)
```
mperrin: Slight suggested revision here: 
```
if title is None:
    title = default_title + " for " + self.name
plt.title(title)
```

mperrin:
incidentally who filed the original issue in STARS? We should remember that and acknowledge them in the (eventual) release notes for the next version, with thanks for pointing this out.

shanosborne:
The reporter was Erik Johansson with the NSO

Issue #271, opened 2018-12-06
Implement and test PSD class for optical noise


mperrin:
Assigned to Iva Laginja (for optics lab work) but I can't seem to tag here here.

mperrin:
Implemented in PR #315 and #317.

PR #300, opened 2019-03-12
Statistical PSD WFE

ivalaginja:
This PR addresses issue [#271](https://github.com/spacetelescope/poppy/issues/271) in creating a WFE class for statistical PSD wavefront errors.

### DIFF
--- a/poppy/wfe.py
+++ b/poppy/wfe.py
@@ -331,27 +331,37 @@ class SineWaveWFE(WavefrontError):
 class StatsticalPSDWFE(WavefrontError):
     """
     Statistical PSF WFE class from power law for optical noise.
-
-    Input WFE in radians RMS.
     """
 
     @utils.quantity_input(wfe=u.rad)
-    def __init__(self, name='PSD WFE', index=3.0, wfe=0.25, **kwargs):
+    def __init__(self, name='PSD WFE', index=3.0, wfe=0.25, seed=None, **kwargs):
 
         super(WavefrontError, self).__init__(name=name, **kwargs)
         self.index = index
         self.wfe_rad = wfe
+        self.seed = seed
 
-    def get_opd(self, wave, units='meters'):
+    def get_opd(self, wave):
+        """
+        Parameters
+        ----------
+        wave : poppy.Wavefront (or float)
+            Incoming Wavefront before this optic to set wavelength and
+            scale, or a float giving the wavelength in meters
+            for a temporary Wavefront used to compute the OPD.
+        seed : integer
+            seed for the random phase screen generator
+        """
 
         y, x = self.get_coordinates(wave)
         rho, theta = _wave_y_x_to_rho_theta(y, x, self.radius.to(u.meter).value)
         psd = np.power(rho, -self.index)
         psd = psd / np.sum(psd) * np.square(self.wfe_rad / wave.wavelength.to(u.meter).value)
 
-        rndm_phase = np.random.random((len(y), len(x)))
+        np.random.seed(self.seed)
+        rndm_phase = np.random.normal(size=(len(y), len(x)))
         rndm_psd = np.fft.fftshift(np.fft.fft2(np.fft.fftshift(rndm_phase)))
         scaled = np.sqrt(psd) * rndm_psd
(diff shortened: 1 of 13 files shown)


poppy/wfe.py
```diff
@@ -323,3 +324,32 @@ def get_opd(self, wave, units='meters'):
...
+
+        super(WavefrontError, self).__init__(name=name, **kwargs)
+        self.index = index
+        self.wfe_rad = wfe
+
+    def get_opd(self, wave, units='meters'):
+
+        y, x = self.get_coordinates(wave)
+        rho, theta = _wave_y_x_to_rho_theta(y, x, self.radius.to(u.meter).value)
+        psd = np.power(rho, -self.index)
```
ivalaginja: I need to add setting this to zero at the center for piston removal.

poppy/wfe.py
```diff
@@ -323,3 +324,32 @@ def get_opd(self, wave, units='meters'):
...
+        psd = psd / np.sum(psd) * np.square(self.wfe_rad / wave.wavelength.to(u.meter).value)
+
+        rndm_phase = np.random.random((len(y), len(x)))
```
ivalaginja: Gauss distributed samples would be better here; plus adding a seed for the random number generator to be able to regenerate exactly the same phase screens.

poppy/wfe.py
```diff
@@ -323,3 +324,32 @@ def get_opd(self, wave, units='meters'):
...
+        rndm_psd = np.fft.fftshift(np.fft.fft2(np.fft.fftshift(rndm_phase)))
+        scaled = np.sqrt(psd) * rndm_psd
+        phase_screen = np.real_if_close(np.fft.ifftshift(np.fft.ifft2(np.fft.ifftshift(scaled))))
```
ivalaginja: This needs a proper normalization.

poppy/wfe.py
```diff
@@ -323,3 +324,32 @@ def get_opd(self, wave, units='meters'):
...
             opd /= wave.wavelength.to(u.meter).value
         return opd
+
+
+class StatsticalPSDWFE(WavefrontError):
+    """
+    Statistical PSF WFE class from power law for optical noise.
+
+    Input WFE in radians RMS.
+    """
```
mperrin: Doc string should define what the input parameters are. 

(Note that poppy uses the numpy docstring convention, rather than the ReST docstring convention that hicat-package uses. See other poppy classes for examples)

poppy/wfe.py
mperrin: Since this part is somewhat cryptic and math-y, it might help to add citations of any relevant reference doc for the equations you used in implementing this, or else acknowledge and give credit to any other package you adapted this from, etc.

poppy/wfe.py
mperrin: Agreed. Seed could be a parameter passed in to the init function of the class.

poppy/wfe.py
```diff
@@ -323,3 +324,32 @@ def get_opd(self, wave, units='meters'):
...
+
+    @utils.quantity_input(wfe=u.rad)
+    def __init__(self, name='PSD WFE', index=3.0, wfe=0.25, **kwargs):
```
mperrin: It seems problematic to have the wfe parameter in units of radians, since that must be relative to some given wavelength. And below you use the wave.wavelength variable for normalizing this. If you performed a broadband calculation on this, the effect would be that the WFE (in units of nm rms) gets larger as the wavelength gets longer -  But if we think about some real mirror with polishing residuals, the total WFE on it is constant independent of wavelength. So it seems to me we need to think a bit more about how to define this normalization parameter.

poppy/wfe.py
mperrin: Also, what is the benefit of calling `np.real_if_close` if you're just going to return `phase_screen.real` on the next line? Is that redundant?

poppy/wfe.py
ivalaginja: @mperrin Is it really necessary here to explicitly set the radial array to zero in the center? With an odd-pixeled wave object, the function _wave_y_x_to_rho_theta() takes care of that anyway as far as I can see, and with an even-pixeled array there is no center pixel.

mperrin:
We're replacing this PR with a new one with a cleaner commit history.

PR #304, opened 2019-03-21
Use RectBivariateSpline instead of zoom for detector-plane resampling

sdwill:
The current implementation of detector-plane resampling for Fresnel propagations uses `scipy.ndimage.zoom`, which appears to result in incorrect amplitude scaling and PSF centering.  This PR replaces this with an alternative implementation which uses `scipy.interpolate.interp2d`, which is better-behaved in both of these aspects.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -1057,7 +1057,7 @@ class FresnelOpticalSystem(BaseOpticalSystem):
         self.distances = []  # distance along the optical axis to each successive optic
 
     @u.quantity_input(distance=u.m)
-    def add_optic(self, optic=None, distance=0.0 * u.m):
+    def add_optic(self, optic=None, distance=0.0 * u.m, index=None):
         """ Add an optic to the optical system
 
         Parameters
@@ -1066,9 +1066,19 @@ class FresnelOpticalSystem(BaseOpticalSystem):
             Some optic
         distance : astropy.Quantity of dimension length
             separation distance of this optic relative to the prior optic in the system.
+        index : int
+            Index at which to insert the new optical element
+
         """
-        self.planes.append(optic)
-        self.distances.append(distance.to(u.m))
+        if index is None:
+            # Optic is appended to the end of the system
+            self.planes.append(optic)
+            self.distances.append(distance.to(u.m))
+        else:
+            # Insert the optic into the middle of the beam train somewhere
+            self.planes.insert(index, optic)
+            self.distances.insert(index, distance.to(u.m))
+
         if self.verbose:
             _log.info("Added optic: {0} after separation: {1:.2e} ".format(self.planes[-1].name, distance))
 
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -971,6 +971,7 @@ class FresnelWavefront(BaseWavefront):
         if np.abs(pixscale_ratio - 1.0) < 1e-3:
             _log.debug("Wavefront is already at desired pixel scale "
                        "{:.4g}.  No resampling needed.".format(self.pixelscale))
+            self.wavefront = utils.pad_or_crop_to_shape(self.wavefront, detector.shape)
             return
 
         super(FresnelWavefront, self)._resample_wavefront_pixelscale(detector)
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -67,7 +67,7 @@ class QuadPhase(poppy.optics.AnalyticOpticalElement):
         k = 2 * np.pi / wave.wavelength.to(u.m).value
         if (z == np.inf) | (z == -np.inf):
             lens_phasor = 1 + 0j
-            _log.debug("lens_phasor:"+str(lens_phasor))
+            _log.debug("lens_phasor:" + str(lens_phasor))
             return lens_phasor
         if accel_math._USE_NUMEXPR:
             rsqd = ne.evaluate("(x ** 2 + y ** 2)")
@@ -435,7 +435,7 @@ class FresnelWavefront(BaseWavefront):
             pixel_scale_x, pixel_scale_y = pixelscale_mpix, pixelscale_mpix
 
         if accel_math._USE_NUMEXPR:
-            return ne.evaluate("pixel_scale_y * y"),  ne.evaluate("pixel_scale_x * x")
+            return ne.evaluate("pixel_scale_y * y"), ne.evaluate("pixel_scale_x * x")
         else:
             return pixel_scale_y * y, pixel_scale_x * x
 
@@ -586,7 +586,7 @@ class FresnelWavefront(BaseWavefront):
             self.planetype = PlaneType.image  # needed for back compatibility when using image plane optics
         elif optic.planetype == PlaneType.detector:
             self._resample_wavefront_pixelscale(optic)
-            self.location = 'at detector '+optic.name
+            self.location = 'at detector ' + optic.name
         else:
             self.location = 'before ' + optic.name
 
@@ -629,7 +629,7 @@ class FresnelWavefront(BaseWavefront):
         x, y = self.coordinates()  # meters
         meter_per_pix = self.pixelscale.to(u.m / u.pix).value
         rhosqr = accel_math._fftshift((x / (meter_per_pix ** 2 * self.n)) ** 2 + (
-                                  y / (meter_per_pix ** 2 * self.n)) ** 2)
+            y / (meter_per_pix ** 2 * self.n)) ** 2)
         # Transfer Function of diffraction propagation eq. 22, eq. 87
         wavelen_m = self.wavelength.to(u.m).value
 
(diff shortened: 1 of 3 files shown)


poppy/fresnel.py
```diff
@@ -1066,16 +1067,27 @@ def add_optic(self, optic=None, distance=0.0 * u.m):
             Some optic
         distance : astropy.Quantity of dimension length
             separation distance of this optic relative to the prior optic in the system.
+        index : int
+            Index at which to insert the new optical element
+
         """
-        self.planes.append(optic)
-        self.distances.append(distance.to(u.m))
+        if index is None:
```
mperrin: I could have sworn we already had this `index` parameter addition in some other PR already. Am I misremembering? Or was that just in the Fraunhofer code? 

In any case this is clearly good to have.

poppy/poppy_core.py
```diff
@@ -199,7 +199,7 @@ def __mul__(self, optic):
 
     def __iadd__(self, wave):
         """Add another wavefront to this one"""
-        if not isinstance(wave, Wavefront):
+        if not isinstance(wave, BaseWavefront):
```
mperrin: Unrelated to the main point of this PR, but looking at this function makes me realize it needs a bit more careful handling. It should only be possible to add wavefronts of the same class together. And with the same pixel scale, and wavelength, and at the same plane type if Fraunhofer...

poppy/poppy_core.py
```diff
@@ -687,16 +699,31 @@ def _resample_wavefront_pixelscale(self, detector):
...
-        # That makes a difference if the detector is only sampling a small part of a much larger wavefront
-
-        new_wf_real = scipy.ndimage.zoom(self.wavefront.real, pixscale_ratio, order=detector.interp_order)
-        new_wf_imag = scipy.ndimage.zoom(self.wavefront.imag, pixscale_ratio, order=detector.interp_order)
-        new_wf = new_wf_real + 1.j*new_wf_imag
+        def make_axis(npix, step):
+            """ Helper function to make coordinate axis for interpolation """
+            return step * np.arange(-npix // 2, npix // 2, dtype=np.float64)
+
+        # Provide 2x margin around image to reduce interpolation errors at edge, but also make
```
mperrin: Can  you explain this 2x margin factor more? That seems a somewhat blunt approach, and I expect would make the interpolation on the order of 4x slower?

poppy/poppy_core.py
```diff
@@ -687,16 +699,31 @@ def _resample_wavefront_pixelscale(self, detector):
...
+        # sure that image is centered properly after it gets cropped down to detector size
+        margin_factor = 2
+        x_in = make_axis(self.wavefront.shape[0], self.pixelscale)
+        y_in = make_axis(self.wavefront.shape[1], self.pixelscale)
+        x_out = make_axis(margin_factor * detector.shape[0], detector.pixelscale)
+        y_out = make_axis(margin_factor * detector.shape[1], detector.pixelscale)
+
+        def interpolator(arr):
+            """ Bind arguments to scipy's interp2d function """
+            # TODO: scipy.interpolate.RectBivariateSpline is much more efficient for data on a
```
mperrin: Would it make sense to try to address this `TODO` idea prior to merging this in?

poppy/fresnel.py
sdwill: Yes, I believe this was merged previously.  I'm not sure why this commit was group in with this PR again.

poppy/poppy_core.py
sdwill: We definitely want to provide some margin around the image, so I chose a small integer multiple to guarantee that there wouldn't be any issues with PSF centering after cropping.

This can be made much smaller, as long as the margin is an integer number of pixels on all sides.  I don't have a good sense of how large the margin needs to be to minimize interpolation edge artifacts- maybe a small number of pixels is enough.

mperrin:
Thanks for those latest commits - waiting now for the Travis tests to complete. 

We should perhaps also think about improving the `test_detector_in_fresnel_system` unit test function - right now it verifies only some very basic parameters like the output shape and pixel scale, but doesn't actually check the contents of the arrays.  Perhaps we should add computation of a radial profile there and comparison to an Airy function, and a check of the flux normalization, etc. Those would go a long way to demonstrating we no longer have incorrect amplitude scaling and PSF centering... I might have a brief bit of time to work on this today.

poppy/poppy_core.py
```diff
@@ -687,21 +687,42 @@ def _resample_wavefront_pixelscale(self, detector):
...
+        # Crop wavefront down to detector size + margin- don't waste computation interpolating
+        # parts of plane that get cropped out later anyways
+        cropped_wf = utils.pad_or_crop_to_shape(self.wavefront, crop_shape)
+
+        # Input and output axes for interpolation.  The interpolated wavefront will be evaluated
+        # directly onto the detector axis, so don't need to crop afterwards.
+        x_in = make_axis(crop_shape[0], self.pixelscale)
+        y_in = make_axis(crop_shape[1], self.pixelscale)
+        x_out = make_axis(detector.shape[0], detector.pixelscale)
+        y_out = make_axis(detector.shape[1], detector.pixelscale)
```
mperrin: Hi @sdwill, in my testing I found one issue here. The pixel scales need be converted to the same units here - this covers the case where e.g. the detector pixel scale is specified in microns/pixel but the wavefront is in meters/pixel. 

```
        x_in = make_axis(crop_shape[0], self.pixelscale.to(u.m/u.pix).value)
        y_in = make_axis(crop_shape[1], self.pixelscale.to(u.m/u.pix).value)
        x_out = make_axis(detector.shape[0], detector.pixelscale.to(u.m/u.pix).value)
        y_out = make_axis(detector.shape[1], detector.pixelscale.to(u.m/u.pix).value)
```

sdwill: That's fair, and I admit I overlooked it during implementation.

mperrin:
> We should perhaps also think about improving the test_detector_in_fresnel_system unit test function - right now it verifies only some very basic parameters like the output shape and pixel scale, but doesn't actually check the contents of the arrays.

I wrote a better version of this test. I can either send it to you to be added into this PR, or more easily, just merge it in myself separately after we complete this PR.

sdwill:
Go ahead and send it to me, and I'll include it with the PR.

mperrin:
```python
def test_detector_in_fresnel_system(npix=256):
    """ Show that we can put a detector in a FresnelOpticalSystem
    and it will resample the wavefront to the desired sampling and size"""

    output_npix = 400
    out_pixscale = 210

    # Setup Fresnel system, with a detector that changes the sampling
    # note - ensure pupil array diameter is at least a bit larger than the actual aperture
    osys = fresnel.FresnelOpticalSystem(pupil_diameter=0.051*u.m, npix=npix, beam_ratio=0.25)
    osys.add_optic(optics.CircularAperture(radius=0.025))
    osys.add_optic(optics.ScalarTransmission(), distance=10*u.m)
    osys.add_detector(pixelscale=out_pixscale*u.micron/u.pixel, fov_pixels=output_npix)

    # Calculate a PSF
    psf, waves = osys.calc_psf(wavelength=1e-6, return_intermediates=True)

    # Check the output pixel scale is as desired
    np.testing.assert_almost_equal(psf[0].header['PIXELSCL'],  out_pixscale/1e6)

    # Check the wavefront gets cropped to the right size of pixels, from something different
    assert waves[0].shape == (1024, 1024)
    assert waves[1].shape == (1024, 1024)
    assert waves[2].shape == (output_npix, output_npix)
    assert psf[0].data.shape == (output_npix, output_npix)

    assert psf[0].header['NAXIS1'] == output_npix


    # Check the PSF is centered and not offset
    psfdata = psf[0].data
    ny, nx  = psfdata.shape
    assert psfdata[ny//2,nx//2]==psfdata.max(), "Peak (spot of Arago) is not in the center"
    assert np.allclose(psfdata[output_npix//2],
                       np.roll(psfdata[output_npix//2][::-1],1), atol=3e-8), "PSF is unexpectedly asymmetric"
    y, x = np.indices(psfdata.shape)
    x-=nx//2
    y-=ny//2
    assert (psf[0].data*x).sum()/psf[0].data.sum() < 0.001, "PSF is surprisingly offset from centered in X"
    assert (psf[0].data*y).sum()/psf[0].data.sum() < 0.001, "PSF is surprisingly offset from centered in Y"

    # Check flux conservation
    # Note this test relies on the detector covering a sufficiently large area of the PSF that the
    # encircled energy is nearly total
    assert np.abs(waves[2].total_intensity-1)< 0.001, "PSF total flux is surprisingly different from 1"
```

mperrin:
@sdwill I think this is ready to merge now, would you agree? 

(For the record I'm not concerne with the coveralls "error" above; we've started getting weird/inconsistent results from Coveralls but haven't yet migrated to a new test coverage service. Despite the "-1.3%" message, when I click through to Coveralls itself it shows zero new lines with missed test coverage. 🤷‍♂️ )

sdwill:
Yes, I'm all for merging.  I've tested this informally against my own wavefront control simulations and it works as expected.

PR #306, opened 2019-03-22
Add FITSOpticalElement support for wavelength-independent phase maps

### LINKED ISSUE #305: BUNIT of 'radian' for FITSOpticalElement
While implementing our fancy vector apodizing phase plate in a POPPY model, I realized there's no obvious way to define a FITSOpticalElement in terms of phase alone. Because vAPPs  use "vector phase", the phase map is actually defined in radians / waves independent of wavelength.

If there's interest, I'd like to add support for phase maps in radians for FITSOpticalElements. Of course, I could just subclass it myself, but in the spirit of open-source I thought I'd see if you wanted this upstream 😄

joseph-long:
Certain optics impart a phase which is (approximately) wavelength independent, using a phenomenon sometimes called Pancharatnam-Berry phase [1,2]. This adds support for such optics in FITSOpticalElement with a new `BUNIT` keyword value, `'radian'`, which causes POPPY to treat the phase map as the same number of radians regardless of wavelength.

(This is currently implemented by scaling the phase map to OPD, only to re-scale back to phase in `get_phasor`. This seemed like the best way to preserve backwards-compatibility.)

Fixes #305 

[1] Pancharatnam, S., “Generalized Theory of Interference, and Its Applications. Part I. Coherent Pencils,” Proc. In- dian. Acad. Sci. Sect. A 44, 247 (1956).

[2] Berry, M., “The Adiabatic Phase and Pancharatnams Phase for Polarized Light,” J. Mod. Opt. 34, 1401–1407 (1987).

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -2418,9 +2418,9 @@ class OpticalElement(object):
             halfsize = 1.0
         extent = [-halfsize, halfsize, -halfsize, halfsize]
 
-
-        ampl = self.get_transmission(wavelength)
-        opd = self.get_opd(wavelength)
+        temp_wavefront = Wavefront(wavelength)
+        ampl = self.get_transmission(temp_wavefront)
+        opd = self.get_opd(temp_wavefront)
         opd[np.where(ampl == 0)] = np.nan
 
         if what == 'both':
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -2634,7 +2634,7 @@ class FITSOpticalElement(OpticalElement):
         self.amplitude_file = None
         self.amplitude_header = None
         self.opd_header = None
-        self._opd_is_phase_map = False
+        self._opd_in_radians = False
         self.planetype = planetype
 
         _log.debug("Trans: " + str(transmission))
@@ -2754,14 +2754,14 @@ class FITSOpticalElement(OpticalElement):
             elif opdunits in ('nanometer', 'nm'):
                 self.opd *= 1e-9
             elif opdunits == 'radian':
-                self._opd_is_phase_map = True
+                self._opd_in_radians = True
             else:
                 raise ValueError(
                     "Got opdunits (or BUNIT header keyword) {}. Valid options "
                     "are meter, micron, nanometer, or radian.".format(repr(opdunits))
                 )
 
-            if self.opd_header is not None and not self._opd_is_phase_map:
+            if self.opd_header is not None and not self._opd_in_radians:
                 self.opd_header['BUNIT'] = 'meter'
 
             if len(self.opd.shape) != 2 or self.opd.shape[0] != self.opd.shape[1]:
@@ -2929,7 +2929,7 @@ class FITSOpticalElement(OpticalElement):
             wavelength = wave.wavelength
         else:
             wavelength = wave
-        if self._opd_is_phase_map:
+        if self._opd_in_radians:
             return self.opd * wavelength.to(u.m) / (2 * np.pi)
         return self.opd
 
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -2289,11 +2289,13 @@ class OpticalElement(object):
             else:
                 # raise NotImplementedError("Need to implement resampling.")
                 zoom = (self.pixelscale / wave.pixelscale).decompose().value
-                resampled_opd = scipy.ndimage.interpolation.zoom(self.opd, zoom,
-                                                                 output=self.opd.dtype,
+                original_opd = self.get_opd(wave)
+                resampled_opd = scipy.ndimage.interpolation.zoom(original_opd, zoom,
+                                                                 output=original_opd.dtype,
                                                                  order=self.interp_order)
-                resampled_amplitude = scipy.ndimage.interpolation.zoom(self.amplitude, zoom,
-                                                                       output=self.amplitude.dtype,
+                original_amplitude = self.get_transmission(wave)
+                resampled_amplitude = scipy.ndimage.interpolation.zoom(original_amplitude, zoom,
+                                                                       output=original_amplitude.dtype,
                                                                        order=self.interp_order)
                 _log.debug("resampled optic to match wavefront via spline interpolation by a" +
                            " zoom factor of {:.3g}".format(zoom))
@@ -2632,7 +2634,7 @@ class FITSOpticalElement(OpticalElement):
         self.amplitude_file = None
         self.amplitude_header = None
         self.opd_header = None
-
+        self._opd_is_phase_map = False
         self.planetype = planetype
 
         _log.debug("Trans: " + str(transmission))
@@ -2734,7 +2736,7 @@ class FITSOpticalElement(OpticalElement):
                 try:
                     opdunits = self.opd_header['BUNIT']
                 except KeyError:
-                    _log.error("No opdunit keyword supplied, and BUNIT keyword not found in header. "
+                    _log.error("No opdunits keyword supplied, and BUNIT keyword not found in header. "
                                "Cannot determine OPD units")
                     raise Exception("No opdunit keyword supplied, and BUNIT keyword not found in header. "
                                     "Cannot determine OPD units.")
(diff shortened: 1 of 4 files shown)


joseph-long:
Interestingly, the test failure is in part of the test using POPPY in ways that it already supported...

```
___________________ test_OPD_in_waves_for_FITSOpticalElement ___________________
    def test_OPD_in_waves_for_FITSOpticalElement():
        pupil_radius = 1 * u.m
        pupil = poppy.CircularAperture(radius=pupil_radius)
        reference_wavelength = 1 * u.um
        single_wave_1um_lens = poppy.ThinLens(
            name='Defocus',
            nwaves=1,
            reference_wavelength=reference_wavelength,
            radius=pupil_radius
        )
        osys = poppy.OpticalSystem(oversample=1, npix=1024)
        osys.add_pupil(pupil)
        osys.add_pupil(single_wave_1um_lens)
        osys.add_detector(0.01 * u.arcsec / u.pixel, fov_pixels=3)
        # We have applied 1 wave of defocus at 1 um, so verify the center
        # has lower flux than at 2 um (it should be the 'hole' of the donut)
        psf_1um = osys.calc_psf(reference_wavelength)
        center_pixel_value = psf_1um[0].data[1,1]
        psf_2um = osys.calc_psf(reference_wavelength)
>       assert psf_2um[0].data[1,1] > psf_1um[0].data[1,1]
E       assert 1.9512451838487682e-11 > 1.9512451838487682e-11
```

poppy/poppy_core.py
```diff
@@ -2632,7 +2634,7 @@ def __init__(self, name="unnamed optic", transmission=None, opd=None, opdunits=N
         self.amplitude_file = None
         self.amplitude_header = None
         self.opd_header = None
-
+        self._opd_is_phase_map = False
```
mperrin: would `._opd_in_radians` be clearer? That more directly connects it to the relevant input parameter.

joseph-long: Yeah, I suppose so. If we wanted to add OPD in waves later, we could rename it since it's an internal attribute 🙃

poppy/poppy_core.py
```diff
@@ -2289,11 +2289,13 @@ def get_phasor(self, wave):
             else:
                 # raise NotImplementedError("Need to implement resampling.")
                 zoom = (self.pixelscale / wave.pixelscale).decompose().value
-                resampled_opd = scipy.ndimage.interpolation.zoom(self.opd, zoom,
-                                                                 output=self.opd.dtype,
+                original_opd = self.get_opd(wave)
```
mperrin: There are also some direct uses of `self.opd` in the `display()` function, which I think would similarly need to be changed to calls to `get_opd` to behave correctly with this kind of input.

poppy/tests/test_core.py
```diff
@@ -498,6 +498,43 @@ def test_FITSOpticalElement(tempdir='./'):
...
+    psf_2um = osys.calc_psf(2 * reference_wavelength)
+    assert psf_2um[0].data[1,1] > psf_1um[0].data[1,1]
+    # Now, use the single_wave_1um_lens optic to make a
+    # wavelength-independent 1 wave defocus
+    lens_as_fits = single_wave_1um_lens.to_fits(what='opd', npix=1536)
+    lens_as_fits[0].header['BUNIT'] = 'radian'
+    lens_as_fits[0].data *= 2 * np.pi / reference_wavelength.to(u.m).value
+    thin_lens_wl_indep = poppy.FITSOpticalElement(opd=lens_as_fits, opdunits='radian')
+    central_pixels = []
+    for prefactor in (0.5, 1.0, 2.0):
```
mperrin: What's the logic of this part of the test function? You loop over some refactors but there's not any assertions inside the loop

mperrin: Oh I see, you've building a list and check they are all close afterwards

joseph-long: Yeah, no reason not to check within the loop I suppose

poppy/tests/test_core.py
```diff
@@ -498,6 +498,43 @@ def test_FITSOpticalElement(tempdir='./'):
...
+        osys = poppy.OpticalSystem(oversample=1, npix=1024)
+        osys.add_pupil(pupil)
+        osys.add_pupil(thin_lens_wl_indep)
+        osys.add_detector(prefactor * 0.01 * u.arcsec / u.pixel, fov_pixels=3)
+        psf = osys.calc_psf(wavelength=prefactor * u.um)
+        central_pixels.append(psf[0].data[1,1])
+        yc, xc = psf[0].data.shape[0] // 2, psf[0].data.shape[1] // 2
```
mperrin: But I'm not sure this line with `yc, xc` has any effect?

joseph-long: Ah yeah, that's an artifact of when I wasn't doing a 3 pixel FOV

poppy/tests/test_core.py
```diff
@@ -498,6 +498,43 @@ def test_FITSOpticalElement(tempdir='./'):
...
+    osys.add_detector(0.01 * u.arcsec / u.pixel, fov_pixels=3)
+    # We have applied 1 wave of defocus at 1 um, so verify the center
+    # has lower flux than at 2 um (it should be the 'hole' of the donut)
+    psf_1um = osys.calc_psf(reference_wavelength)
+    center_pixel_value = psf_1um[0].data[1,1]
+    psf_2um = osys.calc_psf(2 * reference_wavelength)
+    assert psf_2um[0].data[1,1] > psf_1um[0].data[1,1]
+    # Now, use the single_wave_1um_lens optic to make a
+    # wavelength-independent 1 wave defocus
+    lens_as_fits = single_wave_1um_lens.to_fits(what='opd', npix=1536)
```
mperrin: Any reason for using 1536 here when the above example uses 1024? Probably not critical but it's worth trying to keep the unit tests lightweight while still being thorough. (e.g. 3 pixel FOV; but pupil sampling also drives computation time of course)

joseph-long: It's a magic number to make the test pass, of course! If it's 1024, you get the optic diameter sampled across 1024 pixels, whereas the npix for the optical system is something (subtly) different. If I have it as 1024, the first half (with ThinLens) doesn't agree with the trial in the second half to numerical accuracy.

poppy/poppy_core.py
joseph-long: I grepped for `self.opd` but I don't see what you mean upon a cursory inspection...

mperrin: https://github.com/spacetelescope/poppy/blob/master/poppy/poppy_core.py#L2417

joseph-long: Ok, just pushed a commit for that. @douglase just showed up in Tucson and he says hi 😄

mperrin:
Oddly, the tests appear to be passing locally for me, despite failing on Travis. Not sure what's going on. 

Also, in your latest commit 510fb17, it really shouldn't be necessary to turn the input wavelengths into plain values in meters rather than astropy Quantities. Use of Quantities with units for that parameter is definitely supported. See `test_unit_conversions()` also in test_core.py for the test that verifies that functionality.

mperrin:
Aha, got it. The failure is only triggered if running without `numexpr` (or rather, the `numexpr` branch path handles it gracefully while the pure python path raises an error). 

The actual bug is in the `FITSOpticalElement.get_opd` function. This line:
```python
            return self.opd * wavelength.to(u.m) / (2 * np.pi)
```
should be
```python
            return self.opd * wavelength.to(u.m).value / (2 * np.pi)
```
So that the OPD is handed back in units of meters, but as a plain ndarray rather than one with units. 
With that fix applied, I can back out the unnecessary changes in the prior commit  510fb17, and everything still works as expected, for the execution paths both with and without numexpr.

poppy/poppy_core.py
```diff
@@ -2896,6 +2910,33 @@ def pupil_diam(self):
...
+        --------
+        ndarray giving OPD in meters
+
+        """
+        if isinstance(wave, BaseWavefront):
+            wavelength = wave.wavelength
+        else:
+            wavelength = wave
+        if self._opd_in_radians:
+            return self.opd * wavelength.to(u.m) / (2 * np.pi)
```
mperrin: This line needs to be ```return self.opd * wavelength.to(u.m).value / (2 * np.pi)```

mperrin:
All right, tests pass. Looks good and ready to merge. 

(For the record I'm not so concerned about the stated decrease in test coverage from coveralls - we've started to get inconsistent/weird results from coveralls lately, but have not yet migrated to another test runner. )

Issue #310, opened 2019-04-11
Problem with building Fresnel Optical system with ZernikeWFE in place


jlumbres:
I'm simulating a Fresnel Optical system with some aberration in place (such as a tilted source) using the ZernikeWFE function. However, I'm unable to build the wavefronts when I use calc_psf. It seems there's a problem in calc_psf to take in the FresnelWavefront? 

The Fresnel propagator with the ZernikeWFE has worked for me in the past with version 0.6, but I've since updated to the recent developer version ('0.8.1dev1324.dev1324'). I'm opening this error since I've talked with @douglase on this issue.

Here's sample code I generated:

```
# Make a basic optical system
EP_diam = 10e-3 * u.m
oversamp = 0.125
samp = 256
WAVELEN = 632.8e-9*u.m
fl_OAP = 0.2*u.m

# Initialize FresnelOpticalSystem
testbed = poppy.FresnelOpticalSystem(pupil_diameter = EP_diam, npix=samp, 
                                     beam_ratio=oversamp)

# First surface is the entrance aperture.
testbed.add_optic(poppy.CircularAperture(radius=EP_diam/2))

# Add in some wavefront error from source
aberr_coeff_sequence = [0, 0, 0, 0.9]
# Declare LGS source for testbed
sys_aberr = poppy.ZernikeWFE(radius=EP_diam.value/2, coefficients=aberr_coeff_sequence)
testbed.add_optic(sys_aberr)

# OAP1
testbed.add_optic(poppy.QuadraticLens(fl_OAP, name='OAP-1'))
testbed.add_optic(poppy.CircularAperture(radius=0.025*u.m,name="OAP1 aperture"))

# Focal plane
testbed.add_optic(poppy.ScalarTransmission(planetype=PlaneType.intermediate, 
                                           name="focal plane"), distance=fl_OAP)

fp_psf, fp_wfs = testbed.calc_psf(wavelength=WAVELEN.value, 
                                        display_intermediates=True, 
                                        return_intermediates=True)
```
But then I get a very long-winded error that does not produce wavefronts. Here's a copy of the error at the end and the aperture I produced before the wavefront error was implemented:
```
ValueError: Argument 'wavelength' to function '__init__' must be a number (not 'Wavefront:
        wavelength = 0.6328 micron
        shape = (2048, 2048)
        sampling = 3.90625e-05 m / pix'), and convertable to units='m'.
```
[...]

PR #311, opened 2019-04-11
Wfe patch

douglase:
H/T to @jlumbres for the advance warning, here's a quick, if inelegant fix of #310.

### DIFF
--- a/poppy/wfe.py
+++ b/poppy/wfe.py
@@ -17,7 +17,7 @@ import numpy as np
 import astropy.units as u
 
 from .optics import AnalyticOpticalElement, CircularAperture
-from .poppy_core import Wavefront, PlaneType
+from .poppy_core import Wavefront, PlaneType, BaseWavefront
 from poppy.fresnel import FresnelWavefront
 
 from . import zernike
@@ -35,7 +35,7 @@ def _accept_wavefront_or_meters(f):
 
     @wraps(f)
     def wrapper(*args, **kwargs):
-        if not ( isinstance(args[1], FresnelWavefront) or isinstance(args[1], Wavefront)):
+        if not isinstance(args[1], BaseWavefront):
             wave = Wavefront(wavelength=args[1])
             new_args = (args[0],) + (wave,) + (args[2:])
             return f(*new_args, **kwargs)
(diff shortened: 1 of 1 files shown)


poppy/wfe.py
```diff
@@ -33,7 +35,7 @@ def _accept_wavefront_or_meters(f):
 
     @wraps(f)
     def wrapper(*args, **kwargs):
-        if not isinstance(args[1], Wavefront):
+        if not ( isinstance(args[1], FresnelWavefront) or isinstance(args[1], Wavefront)):
```
mperrin: I think the right fix here is to use the superclass:  `if not isinstance(args[1], BaseWavefront)`  (along with adding an import of `BaseWavefront` above.)  

@douglase, I can go ahead and make that change myself and we can discard this PR, or you can change the PR for it. Either way is fine with me so just let me know how you'd like to proceed.

mperrin: @douglase did you get a chance to see my comment last week on this? Let me know your preference and I'll work on this today or tomorrow.

douglase: I didn't see it earlier, change made.

douglase:
I'm a little concerned that the newly initialized wavefront might cause a problem if folks want a different wavefront class than the one initialized.

mperrin:
good question. In the normal use case, a Wavefront or FresnelWavefront is handed in explicitly. This fallback back only gets used in the case that someone says e.g. `optic.get_opd(2e-6)`. So in that case it's hard to say "folks want a different case than the one initialized" if they're just providing a wavelength. 

I can't remember which use case(s) motivate still supporting that back compatible path that just takes a wavelength. Maybe the right fix here is deprecating that entirely? But clearly it gets used somewhere in the execution path that leads to #310.

douglase:
#310 is passing a wavefront, but since the type was not correct it was trying to create a wavefront by passing a wavefront as the wavelength, which returned an error. I expect it is useful to pass a wavelength to an WFE object for illustrative or diagnostic purposes and it might be worth leaving as is for that case, since it's unlikely as user who hasn't initialized a wavefront will care what type the returned wavefront is.

mperrin:
OK, I'm going to go ahead and merge in this PR now. I think I will open up a separate issue that we should consider removing/deprecating this decorator function - it's not consistent with the rest of the API to be able to pass wavelengths and have it automatically turn into a wavefront, and I can't remember what motivated this to begin with.

Issue #309, opened 2019-04-03
Broken display() method for HexSegmentedDeformableMirror() and MultiHexagonAperture()


ivalaginja:
I'm opening this issue after talking about this to @mperrin.

I wanted to display the OPD of a `HexSegmentedDeformableMirror()` object, but the result is a blank figure:

```py
iris = poppy.dms.HexSegmentedDeformableMirror(name='Iris DM',
                                              rings=2,
                                              flattoflat=1.44 * u.mm,
                                              gap=10 * u.micron)
iris.display(what="both")   # this does not work
plt.show()
```
<img width="748" alt="Screen Shot 2019-04-03 at 15 30 59" src="https://user-images.githubusercontent.com/29508965/55507391-91b75280-5625-11e9-825c-711d888e9054.png">

This is partly true for a `MultiHexagonAperture()` object - they inherit the same `display()` method:
```
ap = poppy.MultiHexagonAperture(rings=2, flattoflat=flat, gap=flat/10)
ap.display()
plt.show()
```
<img width="595" alt="Screen Shot 2019-04-03 at 15 33 41" src="https://user-images.githubusercontent.com/29508965/55507598-fb376100-5625-11e9-8019-76f103a4973d.png">
The transmission is fine, but the OPD should actually be blank here, since it is zero everywhere, not equivalent to the transmission.

When accessing the transmission and/or OPD separately with a wave, this works fine for the `MultiHexagonAperture()` object:
```py
flat = 1*u.m
w = poppy.Wavefront(diam=6*flat, npix=512)
ap = poppy.MultiHexagonAperture(rings=2, flattoflat=flat, gap=flat/10)

plt.figure()
plt.subplot(1, 2, 1)
trans = ap.get_transmission(w)
plt.imshow(trans)  # this does work
plt.title("Transmission")

plt.subplot(1, 2, 2)
opd = ap.get_opd(w)
plt.imshow(opd)
plt.title("OPD")
plt.show()
```
<img width="612" alt="Screen Shot 2019-04-03 at 15 37 59" src="https://user-images.githubusercontent.com/29508965/55507828-8c0e3c80-5626-11e9-8ef3-0b73455e1dba.png">

And also for the `HexSegmentedDeformableMirror()`, except that the aperture within the array scales inversely with its physical size `flat`, which should not be the case and is also not happening in the above `MultiHexagonAperture() case`:
```py
iris = poppy.dms.HexSegmentedDeformableMirror(name='Iris DM',
[...]

PR #314, opened 2019-04-23
Refactor optic display internals to fix display bug #309.

mperrin:
Fixes the bug #309 in display of AnalyticOpticalElements introduced by the changes in #306.  Significant change to how the display works internally for analytic optics --  perhaps overall a simplification. 

Yes, the `AnalyticOpticalElement.display` function goes away entirely; all the relevant functionality has been incorporated into `OpticalElement.display`.  Doing it all in one place seemed the best way to ensure consistent handling of the various cases. 

The basic logic is as follows. 
1. In the `OpticalElement.display`, it checks if the optic has a defined pixelscale sampling (i.e. FITSOpticalElement or ArrayOpticalElement, etc) or if it doesn't (AnalyticOpticalElement and subclasses). Depending on which of those is the case, it sets up an input wavefront with the appropriate sampling for displaying the optic, either relying on the optic's native pixel scale or specifying the requested scale.  (The code for setting up the input wavefront is a redundant copy of code from AnalyticOpticalElement.sample, so that's inelegant, but that seemed the most practical way to achieve this.) 
2. The optic's `get_opd` and `get_transmission` functions are called using that wavefront to evaluate the optic at the requested wavelength and sampling.  
3. The resulting arrays are displayed. I refactored this part to avoid having to recursively call itself for displaying both opd and intensity by putting the core display logic in an internal helper function. This is a better way of doing this than what I had before, and better achieves the goal of avoiding redundant evaluations of the opd and amplitude. 

I've tested this relatively extensively including re-running the entire `available_optics.ipynb` example notebook. I am reasonably confident this is working well now. But since the display code is not well tested by the unit test cases, I'd appreciate some other eyes on this before I merge it...

attn @joseph-long @ivalaginja @douglase


ivalaginja:
I found a minor problem that I don't think originates in the changes done in this PR but it is a bug in the plotting of a `HexSegmentedDeformableMirror()` instance. When I define it with `center=False`, meaning I omit the center segment, the loop in `get_opd()` still tries to address the full number of segments:

```py
import poppy
import astropy.units as u
import matplotlib.pyplot as plt
%matplotlib inline

iris = poppy.dms.HexSegmentedDeformableMirror(name='Iris DM',
                                              rings=3,
                                              flattoflat=15.*u.m,
                                              gap=2*u.cm,
                                              center=False)
plt.figure(figsize=(16, 8))
iris.display(what='both')
```

Error message:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-3-56c9b836c811> in <module>()
     11                                               center=False)
     12 plt.figure(figsize=(16, 8))
---> 13 iris.display(what='both')
[...]
~/anaconda/envs/pastis/lib/python3.6/site-packages/poppy/dms.py in get_opd(self, wave)
    594             wseg = self._seg_indices[i]
    595             self.opd[wseg] = (self._surface[i, 0] +
--> 596                               self._surface[i, 1] * self._seg_x[wseg] +
    597                               self._surface[i, 2] * self._seg_y[wseg])
    598         return self.opd

IndexError: index 36 is out of bounds for axis 0 with size 36
```

It works fine with `center=True`:
```python
iris = poppy.dms.HexSegmentedDeformableMirror(name='Iris DM',
                                              rings=3,
                                              flattoflat=15.*u.m,
                                              gap=2*u.cm,
                                              center=True)
plt.figure(figsize=(16, 8))
iris.display(what='both')
```

mperrin:
I agree, that looks like a separate bug. Can you file it as a new issue, please?

ivalaginja:
Will do in a sec.

mperrin:
For point 1, about the second plot - I wonder if there was some accidental omission or cut-and-paste error in putting the code in that issue. As written, the call to `ap.display()` doesn't specify what to plot, so it uses the default value which is `"intensity"`.  I don't see how that would ever have produced both plots as shown, unless the code she ran was actually `ap.display(what="both")`.  @ivalaginja can you comment on this? 

For point 2, that's concerning if it still doesn't work. Will have to investigate more.

ivalaginja:
@mperrin you're right, I must have lost that `what` keyword when I was creating the issue. I checked the script I used to build the issue, and I indeed used `ap.display(what="both")`, so this should be fine.

Also, I just realized that I might have worded that part of the issue poorly, since the plot for `ap.display(what="both")` actually **is** zero everywhere, as it should, and the area around it is simply masked out. Apologies for that.

ivalaginja:
For point 2, I just merged poppy/master into my fork and this problem started occurring now there too. Maybe I can do a diff and figure out what causes this.

mperrin:
@shanosborne wrote above: 
> In the 4th and 5th plots, Iva commented that there was an issue with the default scaling and it doesn't appear that was fixed in this PR. You might not have been trying to fix it here, but then we just need to make sure we remember it.

I've just taken another look at this. I think the code is working as intended, and instead the issue is in the scaling code used in the example given in #309. 

That code is as follows, copied here for ease of reference: 
```python
iris = poppy.dms.HexSegmentedDeformableMirror(name='Iris DM',
                                              rings=2,
                                              flattoflat=1.44 * u.mm,
                                              gap=10 * u.micron)
plt.figure()
plt.subplot(1, 3, 1)
flat = 0.001*u.m    # the bigger this number, the smaller the display??
w = poppy.Wavefront(diam=6*flat, npix=512)
trans = iris.get_transmission(w)
plt.imshow(trans)   # this works, but weird sizing
plt.title("flat = 0.001*u.m")

plt.subplot(1, 3, 2)
flat = 0.01*u.m    # the bigger this number, the smaller the display??
w = poppy.Wavefront(diam=6*flat, npix=512)
trans = iris.get_transmission(w)
plt.imshow(trans)   # this works, but weird sizing
plt.title("flat = 0.01*u.m")

plt.subplot(1, 3, 3)
flat = 0.1*u.m    # the bigger this number, the smaller the display??
w = poppy.Wavefront(diam=6*flat, npix=512)
trans = iris.get_transmission(w)
plt.imshow(trans)   # this works, but weird sizing
plt.title("flat = 0.1*u.m")
plt.show()
```


For the 3 plots, the above code increases the `flat` variable, which causes the wavefronts `w` to be created with increasingly larger pixel scale -- and therefore the DM _should_ look smaller in the later plots, relative to the pixel scale. Note that the same DM instance `iris` is used for all the plots, and changing the `flat` variable doesn't affect the size of that DM's segments. In other words the plots are zooming out the spatial scale, without changing the size of the DM, so it makes sense that the DM looks smaller. 

I think this explains what's going on, and there doesn't seem to be any remaining bug in this. @shanosborne, @ivalaginja does that make sense to you?

mperrin:
Having heard no objections, I'm going to merge this now. This is a significant enough display but I don't want to leave it open in master any longer.

PR #315, opened 2019-04-24
Add StatisticalPSDWFE class

ivalaginja:
Replacement for PR #300, this addresses issue #271, implementing a WFE class for statistical PSD WFEs.


poppy/tests/test_wfe.py
```diff
@@ -83,3 +84,88 @@ def test_ParameterizedAberration():
...
+    inv_psd = np.fft.ifftshift(np.fft.ifft2(np.fft.ifftshift(psd_opd)))
+    rad, prof = radial_profile(np.abs(inv_psd) ** 2, center=(int(NPIX/2), int(NPIX/2)))
+
+    # Make comparison slope
+    com = rad[5:] ** (-3)
+    com *= prof[5] / com.max()
+
+    # Verify that the slopes of prof and com are the same
+    del_prof = prof[100] - prof[5]
+    del_com = com[100] - com[0]
```
mperrin: I'm having some trouble following how this works in detail: you make the comparison slow after discarding the first 5 elements, ok, so I see that `com[0]` is scaled to have the same value as `prof[5]`. But then if we could 100 elements later from both, wouldn't that be `prof[105]` rather than 100?

poppy/wfe.py
```diff
@@ -325,3 +326,65 @@ def get_opd(self, wave, units='meters'):
         if units == 'waves':
             opd /= wave.wavelength.to(u.meter).value
         return opd
+
+
+class StatisticalPSDWFE(WavefrontError):
+    """
+    Statistical PSF WFE class from power law for optical noise.
```
mperrin: PSF -> PSD

poppy/wfe.py
```diff
@@ -325,3 +326,65 @@ def get_opd(self, wave, units='meters'):
...
+    """
+
+    @utils.quantity_input(wfe=u.nm, radius=u.meter)
+    def __init__(self, name='PSD WFE', index=3.0, wfe=50*u.nm, radius=1*u.meter, seed=None, **kwargs):
+        """
+        Parameters
```
mperrin: This whole docstring of `__init__` should be appended/merged into the class doc string above.  Python doesn't define what should go into docstrings or where, and as a result there are some competing standards for how to do so. poppy tries to follow the numpy doc string standards, which states that the parameters to `__init__` should be documented in the class docstring. See https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

The point of this is, I suppose, that you can say `help(ClassName)` rather than `help(ClassName.__init__)`. Again this is partially a stylistic thing, but it's good to do for consistency with all the other classes in poppy.

poppy/wfe.py
```diff
@@ -325,3 +326,65 @@ def get_opd(self, wave, units='meters'):
...
+        ----------
+        wave : poppy.Wavefront (or float)
+            Incoming Wavefront before this optic to set wavelength and
+            scale, or a float giving the wavelength in meters
+            for a temporary Wavefront used to compute the OPD.
+        units : 'meters' or 'waves'
+            Coefficients are supplied as meters of OPD, but the
+            resulting OPD can be converted to
+            waves based on the `Wavefront` wavelength or a supplied
+            wavelength value.
```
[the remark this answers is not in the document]
ivalaginja: Gotcha

poppy/wfe.py
```diff
@@ -325,3 +326,65 @@ def get_opd(self, wave, units='meters'):
...
+            negative power law spectra index, defaults to 3
+        wfe: astropy quantity
+            wfe in linear astropy units, defaults to 50 nm
+        radius: astropy quantity
+            radius of optic in linear astropy units, defaults to 1 m
+        seed : integer
+            seed for the random phase screen generator
+        """
+
+        super(WavefrontError, self).__init__(name=name, **kwargs)
```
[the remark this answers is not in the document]
ivalaginja: Changed accordingly.

poppy/tests/test_wfe.py
```diff
@@ -83,3 +84,88 @@ def test_ParameterizedAberration():
 
     np.testing.assert_allclose(pd_wave.phase, zern_wave.phase,
                                err_msg="ParameterizedAberration disagrees with ZernikeAberration")
+
+
+def test_StatistcalPSDWFE():
```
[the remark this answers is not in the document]
ivalaginja: Fixed.

poppy/tests/test_wfe.py
ivalaginja: Very true. I fixed it accordingly.

poppy/wfe.py
ivalaginja: Fixed.

poppy/wfe.py
ivalaginja: True, I've never seen a doctstring in the `__init__` and I have no clue why I put it there. I moved them to the class doctstring.

ivalaginja:
@mperrin I've gone through your review and added the changes. I also included the example in the documentation. Please let me know if there is anything else.

mperrin:
Hmm, something seems to have gotten messed up with the commit diff for the notebook - the version in this PR gives me a JSON formatting error when I try to open it. Specifically the line with the embedded encoding of the PNG image is truncated and missing the last part - I can tell this because it doesn't end with the sequence `=\n",` which all other embedded PNGs end with.  This will need to be fixed before this PR can be merged. 

I also realized there is an issue with the `test_StatisticalPSDWFE` function, specifically this part checking the slope: 
```
# Make comparison slope
com = rad[5:] ** (-3)
com *= prof[5] / com.max()

# Verify that the slopes of prof and com are the same
del_prof = prof[105] - prof[5]
del_com = com[100] - com[0]
```
Because of the spectral index, the values of `prof[105]` or `com[100]` are much lower than that of their initial elements, such that they are basically negligible in comparison. Therefore `del_prof` ~ `prof[5]` and `del_com` ~`com[0]`. And those are normalized to be precisely equal on the previous line. As a result the above test code passes even if we change the slope drastically. I tried this changing the comparison slope to 6 instead of 3, on the first line of the above, and the test still passes anyway. 

The more rigorous way to test this is to actually measure the spectral index / PSD of the output WFE map, then compare it to the requested value. I spent a little while working out how to do this with the astropy modeling framework. It turns out to be somewhat subtle because of the statistical/random nature of this class; no specific instance will exactly match what is required. I also encountered some weirdnesses in getting the astropy model fitter to converge on a good fit. Here's what I came up with in the end: 

```
import astropy.modeling
# guess a power law model; but don't have the starting guess be the right anwser
plaw_guess = astropy.modeling.models.PowerLaw1D(alpha=INDEX*1.5)

# Perform a fit. Drop some initial and trailing values to avoid numerical artifacts. 
fitter = astropy.modeling.fitting.LevMarLSQFitter()
drop = 20  # how many array elements to discard from fit
prof_norm = prof/prof.max() # Empirically this fit process works better if we normalize the profile first.
plaw_fit = fitter(plaw_guess, rad[drop:-drop], prof_norm[drop:-drop], weights=(prof_norm[drop:-drop])**-2)

# check the spectral index is as desired, within at least a few percent
assert np.isclose(INDEX, plaw_fit.alpha, rtol=0.03), ("Measured output spectral "
        "index doesn't match input within 3%: "
        " {} vs {}".format(INDEX, plaw_fit.alpha) )

```
Also, it's necessary to fix the seed in the test. Otherwise we run the risk of it sometimes randomly failing due to an outlier statistical draw. This could be avoided by setting a very wide tolerance, more than the 3% I used above, but it seems better to me to fix the seed to a specific value and demonstrate that a relatively tight test can be passed. I found that the seed `123456` results in a pass for the specific test case of interest here, but that was just from trial and error and is itself a meaningless value. 

@ivalaginja  - at your choice, I can let you add in additional commits for the above, or I can merge this and then go ahead and apply these fixes myself. Your call. Please let me know how you'd like to proceed.

ivalaginja:
I see. Since you know exactly how to do it, please go ahead and apply the changes.

PR #316, opened 2019-04-25
Add new OPD keywords

shanosborne:
Added new keywords `OPD_FILE` and `OPDSLICE` to the FITS header to separate the information previously held just in `PUPILOPD`. I added these keywords to all the different cases (even though it may not be necessary) to try and keep things consistent.

From issue: https://github.com/spacetelescope/webbpsf/issues/297

### DIFF
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -436,15 +436,30 @@ class Instrument(object):
 
         if self.pupilopd is None:
             opdstring = "NONE - perfect telescope! "
+            opdfile = 'None'
+            opdslice = 'None'
         elif isinstance(self.pupilopd, str):
             opdstring = os.path.basename(self.pupilopd)
+            opdfile = self.pupilopd
+            opdslice = 0  # default slice
         elif isinstance(self.pupilopd, fits.HDUList):
             opdstring = 'OPD from supplied FITS HDUlist object'
+            if isinstance(self.pupilopd.filename(), str):
+                opdfile = os.path.basename(self.pupilopd.filename())
+            else:
+                opdfile = 'None'
+            opdslice = 'None'
         elif isinstance(self.pupilopd, poppy_core.OpticalElement):
             opdstring = 'OPD from supplied OpticalElement: ' + str(self.pupilopd)
+            opdfile = str(self.pupilopd)
+            opdslice = 'None'
         else:  # tuple?
             opdstring = "%s slice %d" % (os.path.basename(self.pupilopd[0]), self.pupilopd[1])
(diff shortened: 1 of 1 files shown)


mperrin:
For more consistency, should the OPDSLICE keyword just be 0 instead of None as the default for cases when there's not a data cube?

poppy/instrument.py
```diff
@@ -436,15 +436,30 @@ def _get_fits_header(self, result, options):
...
             opdstring = 'OPD from supplied OpticalElement: ' + str(self.pupilopd)
+            opdfile = str(self.pupilopd)
+            opdslice = 'None'
         else:  # tuple?
             opdstring = "%s slice %d" % (os.path.basename(self.pupilopd[0]), self.pupilopd[1])
+            opdfile = os.path.basename(self.pupilopd[0])
+            opdslice = self.pupilopd[1]
         result[0].header['PUPILOPD'] = (opdstring, 'Pupil OPD source')
+        result[0].header['OPD_FILE'] = (opdfile, 'Pupil OPD file name')
+        result[0].header['OPDSLICE'] = (opdslice, 'Pupil OPD slice number')
```
mperrin: Maybe 'Pupil OPD slice number, if file is a datacube' to make it more clear

poppy/instrument.py
```diff
@@ -436,15 +436,30 @@ def _get_fits_header(self, result, options):
 
         if self.pupilopd is None:
             opdstring = "NONE - perfect telescope! "
+            opdfile = 'None'
+            opdslice = 'None'
```
mperrin: Maybe 0 instead of 'None', for this and all other cases?

poppy/instrument.py
```diff
@@ -436,15 +436,30 @@ def _get_fits_header(self, result, options):
...
         elif isinstance(self.pupilopd, str):
             opdstring = os.path.basename(self.pupilopd)
+            opdfile = self.pupilopd
```
mperrin: For consistency I think this should also be wrapped in `os.path.basename` like the others.

shanosborne:
But doesn't having 0 there still imply that it does come from a datacube (and is the 0th slice)? Whereas None is clear that it's just not an application keyword

mperrin:
> But doesn't having 0 there still imply that it does come from a datacube (and is the 0th slice)? Whereas None is clear that it's just not an application keyword

That's a valid point; but I also don't like a keyword that is sometimes an integer and sometimes a string. It would be nice if the value were of a consistent type.

shanosborne:
Keeping the data type the same does make sense. And I think the addition to the comment also helps make it clear.

PR #317, opened 2019-04-29
amend PR #315 with improved test case and fixed plot in notebook

mperrin:
- `available_optics.ipynb` fixed to avoid the accidentally truncated PNG encoding in PR #315 
 - the `test_StatisticalPSDWFE` unit test was improved to have a more rigorous test that the output PSD index is as requested, and also to be able to display output plots like this: 

```
import poppy.tests.test_wfe as tw
tw.test_StatisticalPSDWFE(plot=True, seed=1234, index=3.4)
```

Screen Shot 2019-04-29 at 2 33 25 PM

attention @ivalaginja


mperrin:
I'm going to decide this one doesn't need a formal code review, because it's just a relatively modest edit to a unit test case of a PR which has already had two sets of eyes on it.  In place of a formal code review, @ivalaginja could you take a brief look at the revised test case code, and let me know if it makes sense to you and the comments are clear?

mperrin:
@ivalaginja, a small reminder for this other PR too. Could you please take a brief look at the revised test case code, and let me know if it makes sense to you and the comments are clear?

ivalaginja:
@mperrin I assume you picked `drop` by visual inspection of the plot and pinpointed where it starts being. more or less regular? I remember that initially we didn't have a strong opinion on how much of the array we wanted to drop when performing the fit.

Otherwise this seems very clear to me!

mperrin:
Yes, the `drop` parameter was picked semi-arbitrarily, to get to where it is more or less regular because of sufficient statistics, yes. Not any deep meaning to that specific value. I think we agree it's not critical, and there's some inevitably looseness needed in any test of a probabilistic function. 

Thanks!

Issue #318, opened 2019-04-30
Indexing error in get_opd() of HexSegmentedDeformableMirror() class when center=False


ivalaginja:
Discovered while testing PR #314 I found a bug in the plotting of a `HexSegmentedDeformableMirror()` instance. When I define it with `center=False`, meaning I omit the center segment, the loop in `get_opd()` still tries to address the full number of segments:

```py
import poppy
import astropy.units as u
import matplotlib.pyplot as plt

iris = poppy.dms.HexSegmentedDeformableMirror(name='Iris DM',
                                              rings=3,
                                              flattoflat=15.*u.m,
                                              gap=2*u.cm,
                                              center=False)
plt.figure(figsize=(16, 8))
iris.display(what='both')
```

Error message:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-3-56c9b836c811> in <module>()
     11                                               center=False)
     12 plt.figure(figsize=(16, 8))
---> 13 iris.display(what='both')
[...]
~/anaconda/envs/pastis/lib/python3.6/site-packages/poppy/dms.py in get_opd(self, wave)
    594             wseg = self._seg_indices[i]
    595             self.opd[wseg] = (self._surface[i, 0] +
--> 596                               self._surface[i, 1] * self._seg_x[wseg] +
    597                               self._surface[i, 2] * self._seg_y[wseg])
    598         return self.opd

IndexError: index 36 is out of bounds for axis 0 with size 36
```

It works fine with `center=True`:
```python
iris = poppy.dms.HexSegmentedDeformableMirror(name='Iris DM',
                                              rings=3,
                                              flattoflat=15.*u.m,
                                              gap=2*u.cm,
                                              center=True)
plt.figure(figsize=(16, 8))
iris.display(what='both')
```

ivalaginja:
This might be easily fixed by adjusting `self.segmentlist` in its base class `MultiHexagonAperture()` depending on whether we have `center=True` or `center=False`. Would need to look into it in more detail though.

mperrin:
Great, if you want to test it out and hopefully submit a PR with a fix that would be great. I expect you are correct and something like that is the necessary fix.

ivalaginja:
Turns out that `self.segmentlist` is behaving as you would expect, it simply omits the `0` if you tell it not to include the center segment.
[...]

PR #319, opened 2019-05-03
Accessing the last segment of a `HexSegmentedDeformableMirror()` instance

ivalaginja:
Addresses Issue #318: in `HexSegmentedDeformableMirror()`, there was no way of talking to the last segment of a segmented primary (e.g. segment no. 18 for `rings=2` or segment no. 36 for `rings=3` etc.). Turns out that is because `self._surface` is an array while `self._seg_indices` is a dictionary. The dictionary keeps track of the segment numbers all right because it is not ordered, but `self._surface` is ordered and hence accesses its entries wrong when asked for a specific segment if the DM has no center segment.

I fixed it by resetting the index for `self._surface` if `center=False`, it's only three lines of code.

It might be worth considering to refactor this such that `self._surface` is also a dictionary. This is also the reason why we developed the IrisAO wavefront code (the class `IrisWavefront()` [here](https://github.com/spacetelescope/iris-ao/blob/controls-jost/controls-jost/DM_functions.py), although I don't know if the link works) to work with segment PTT commands as a dictionary, not an array, since that would mean that the index to access a specific segment is always the same like the segment number itself, which is easier to use. @mperrin, would that make sense?


mperrin:
This fix works, but after thinking about it a little, I think I would like to fix the underlying issue #318 in a different way. This way will achieve the goal you mention above: 
> the index to access a specific segment is always the same like the segment number itself, which is easier to use.

We don't actually need to use a dict for that - we can still use an array. The trick is to allocate array entries for all the segments in all cases, including the 0th center segment even if is disabled. That can be done with a simple change to where `self._surface` is defined in the class `__init__`. Then any iterations over `self.segmentlist` will skip the missing segments. I'll submit this in an alternate PR and you can take a look. 

I think this alternate fix is better because it will keep the indexing more consistent. It also would allow this class to in the future support leaving out an arbitrary list of segments, if desired, which is something the MultiHexagonAperture class can do already via the `segmentlist` parameter.  

(BTW, even though I'm suggesting an alternate fix, thanks very much for this PR. It pinpointed exactly which part of the code I needed to look at, and your suggestion above on indexing is what got me thinking if there was another maybe better way to achieve that goal.)

ivalaginja:
Sounds good!

Issue #199, opened 2018-08-23
unexpected ZernikeWFE.get_opd() behavior


mperrin:
<a href="https://github.com/douglase"><img src="https://avatars1.githubusercontent.com/u/1025951?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [douglase](https://github.com/douglase)**
_Tuesday Nov 22, 2016 at 03:31 GMT_
_Originally opened as https://github.com/mperrin/poppy/issues/199_

----

Kudos to @tukyab for helping find this one.

setup of problem:
```
r=.0254
osys = poppy.OpticalSystem()
osys.add_pupil(poppy.CircularAperture(radius=r))    
osys.add_pupil(poppy.ZernikeWFE(radius=r,coefficients=[0,0,0.000001,0]))   

```

Expected behavior:
```
plt.imshow(osys.planes[1].get_opd(osys.input_wavefront(wavelength=600e-9)))
plt.colorbar()
```
gives:
![unknown](https://cloud.githubusercontent.com/assets/1025951/20509472/ff534a84-b036-11e6-9730-216cf2ec89c3.png)

unexpected behavior:

```
plt.imshow(osys.planes[1].get_opd(600e-9))
plt.colorbar()
```

![unknown-2](https://cloud.githubusercontent.com/assets/1025951/20509474/01360bd4-b037-11e6-810c-dde83c0a45a4.png)

This works when the aperture is a few meters across, but only there, it doesn't rescale at all (if you make r=8 then you can't see the circle anymore), the pixel_scale seems to be getting lost when you just pass a wavelength to get_opd().

The documentation says you can pass a wavelength as a float or a wavefront, but I don't actually see why it doesn't throw an error, since the ```self.circular_aperture.get_transmission(wave)``` function [tests that wave is a wavefront]( https://github.com/mperrin/poppy/blob/master/poppy/optics.py#L848).
@josePhoenix ?

mperrin:
<a href="https://github.com/josePhoenix"><img src="https://avatars3.githubusercontent.com/u/172886?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [josePhoenix](https://github.com/josePhoenix)**
_Tuesday Nov 22, 2016 at 13:39 GMT_

----

Well, that's weird. I'll take a look.

mperrin:
[...]

PR #322, opened 2019-05-14
remove _accept_wavefront_or_meters and units args, and related tests, to fix #312 and #199

### LINKED ISSUE #312: consider deprecating _accept_wavefront_or_meters wrapper in wfe.py
As discussed in comments for #311, in `wfe.py` there is a decorator function `_accept_wavefront_or_meters` which lets you pass either a Wavefront object or a scalar wavelength in meters to the `get_opd` functions etc.  This behavior isn't consistent with the rest of the API, and it doesn't take into account the slight  differences between Wavefront and FresnelWavefront classes. Can't remember now what functionality motivated this. 

It seems simplest and most elegant to just rip this code out, and have those functions behave consistently with everything in `optics.py`: the user must supply a wavefront (i.e. instance of a BaseWavefront subclass) as the function input, or else an exception is raised.

mperrin:
This PR would remove the ability of WFE classes' `get_opd` functions to accept both Wavefront objects or raw floating-point wavelength values . That behavior was inconsistent with how anything else worked, caused problems (see #199), wasn't used anywhere else in poppy, and wasn't described in the documentation. So I think we can simply drop it, for an overall simplification. 

I replaced it with a wrapper that just checks the input argument is a wavefront and raises an error if not. This could be moved upstream into optics.py later for use on all the optic classes there, if it proves sufficiently useful.

### DIFF
--- a/poppy/wfe.py
+++ b/poppy/wfe.py
@@ -28,21 +28,17 @@ __all__ = ['WavefrontError', 'ParameterizedWFE', 'ZernikeWFE', 'SineWaveWFE',
         'StatisticalPSDWFE']
 
 
-def _accept_wavefront_or_meters(f):
+def _check_wavefront_arg(f):
     """Decorator that ensures the first positional method argument
-    is a poppy.Wavefront or a floating point number of meters
-    for a wavelength
+    is a poppy.Wavefront or FresnelWavefront
     """
 
     @wraps(f)
     def wrapper(*args, **kwargs):
         if not isinstance(args[1], BaseWavefront):
-            wave = Wavefront(wavelength=args[1])
-            new_args = (args[0],) + (wave,) + (args[2:])
-            return f(*new_args, **kwargs)
+            raise ValueError("The first argument must be a Wavefront or FresnelWavefront object.")
         else:
             return f(*args, **kwargs)
-
     return wrapper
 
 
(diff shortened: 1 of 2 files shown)


mperrin:
@ivalaginja or @shanosborne would either of you be available to review this PR sometime soon? I forgot to assign any reviewer when I added this PR a week and a half ago.

poppy/wfe.py
```diff
@@ -356,7 +331,7 @@ def __init__(self, name='PSD WFE', index=3.0, wfe=50*u.nm, radius=1*u.meter, see
         self.radius = radius
         self.seed = seed
 
-    @_accept_wavefront_or_meters
+    @_check_wavefront_arg
     def get_opd(self, wave, units='meters'):
```
shanosborne: Did you also want to get rid of the units parameter for this method?

mperrin: oh good catch, I think yes.

mperrin: removed.

shanosborne:
Looks like the docstring and the `if units == 'waves'` part also need to be updated

mperrin:
Agh, right. Shows me not to multitask so much when editing code... Thanks for paying attention and catching when I miss something!

PR #326, opened 2019-05-24
minor: bugfix for pixel scale comparison in adding wavefronts

mperrin:
Trivial bug fix of a float comparison that should be `np.isclose` instead of a strict equality. 

Unrelatedly, also a trivial change to an error string to make the error message clearer.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -267,7 +267,7 @@ class FresnelWavefront(BaseWavefront):
             _log.debug("Skipping oversampling, oversample < 1 or already padded ")
 
         if self.oversample < 2:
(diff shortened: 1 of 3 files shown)


mperrin:
@shanosborne I assigned you as a reviewer but please don't spend more than 5 seconds on this one... :-)

shanosborne:
It looks like that test in failing because now doing `w1+w4` raises `UnitConversionError: 'arcsec / pix' and 'm / pix' are not convertible` from `astropy/units` before it gets a chance to raise the `ValueError` inside `poppy` (coming from line 209 in `poppy_core`)

PR #329, opened 2019-06-18
Improve ContinuousDeformableMirror: subpixel position actuator model, improved coordinate transforms including inclined DMs

mperrin:
This PR improves the ContinuousDeformableMirror class (for e.g. Boston MEMS DMs) by modeling the actuator positions at sub pixel resolution in the "get DM surface via convolution" implementation. 

 - The prior code in master rounds each actuator to the nearest integer pixel, pokes those pixels, then convolves by the DM actuator influence function. This results in uneven spacing of the modeled DM actuators unless the pupil pixel scale is an exact divisor of the DM actuator spacing (which we can't rely on). 
 - The new code distributes the pokes for each DM across a 2x2 square of pixels, linearly weighting between them according to the sub pixel position of the actuator. Then it convolves that by the DM actuator influence function. This results in much more uniform spacing of the modeled actuators, independent of pixel scales in the DM array. 

*Example image before:* 

Unknown

*With this PR:*

Unknown-1

This PR also slightly refactors the handling of DM flips, to reduce duplicated code between the convolution and Gaussian pathways, and to include the DM actuator mask (if provided) in the flipping. This is independent from the main improvements discussed above; a separate small improvement to the DM class.


mperrin:
This pull request also now includes functionality to tilt or incline optics in general (e.g. applies a `cosine(i)` rescaling for an optic that is tilted by `i` degrees with respect to the beam).

mperrin:
Examples of tilted deformable mirrors. Here's a hex dm and continuous dm, both inclined by 45 degrees. 

Unknown

Unknown-1

mperrin:
Note, there is a known issue with the inclination handling, in that it partially confuses the gray pixel approximation code added in PR #325.  That code expects circular apertures rather than elliptical ones, and can emit a warning when used on inclined apertures that the gray pixel values may not be exact. I think this is OK for now - slightly inaccurate gray pixels are still probably better than entirely black and white pixels. Just something to be aware of, and potentially try to improve in the future if it becomes an issue. 

For the specific motivating case of HiCAT, in which the Boston DMs are used pretty close to face on (only inclined by ~ 8º relative to face on), this effect is small enough to be negligible. Not a concern.

mperrin:
Example code for making the above plots: 

```python
hexdm = poppy.HexSegmentedDeformableMirror()
hexdm.inclination_y = 45
hexdm.set_actuator(12, 3e-7, 0, 0)
hexdm.display(what='both')

```
```python
bdm = poppy.ContinuousDeformableMirror(inclination_y=45)
bdm.set_actuator(5, 7, 3e-7)
bdm.set_actuator(1, 4, -2e-7)
bdm.display(what='both')

```

ivalaginja:
This works as expected and the code looks good.

Since there is an implemented warning that it is physically inconsistent to set inclinations on both X and Y at the same time, it might make sense to also add a warning that only a tilt range form -90 to +90 degrees make sense physically, other wise we start hitting the DM back plane. Mathematically, this code works as expected if the DM is modeled as a transmissive mask that acts on the OPD, but physically it might make sense to constrain it.

ivalaginja:
@mperrin I can see you added the warning, but after pulling in the change I can't trigger the warning for some reason. I am running

```python
import poppy
import matplotlib.pyplot as plt

hexdm = poppy.HexSegmentedDeformableMirror()
hexdm.inclination_y = 120
hexdm.set_actuator(12, 3e-7, 0, 0)
hexdm.display(what='both')
plt.show()
```

mperrin:
The check is only done once in the class init. Try
`hexdm = poppy.HexSegmentedDeformableMirror(inclination_y = 120)`
and you will see it. 

(Yes, this is less paranoid than it could be. But at least for now, I'd rather not add this very-rarely-needed check to every single call to `get_coordinates`)

ivalaginja:
It doesn't like that:
```
hexdm = poppy.HexSegmentedDeformableMirror(inclination_y = 120)
TypeError: __init__() got an unexpected keyword argument 'inclination_y'
```

mperrin:
Good catch - the Hex DM class was missing a `**kwargs` passthrough for keyword arguments including all the transformations. Fixed.

mperrin:
Thanks @ivalaginja

PR #343, opened 2019-11-24
Fix RTD builds 

robelgeda:
RTD seems to be building its environment to `readthedocs.yml` and fils to activate that environment. I am not sure exactly why its failing but It seems like we have to rename the environment to match the build which would be a flaw in their end. WebbPSF uses pip file to install its prerequisites which works fine on RTD. I have removed the conda env file in favor of the default pip env file in the docs folder.


robelgeda:
looks like travis does not use the yaml file so we are clear to remove it

robelgeda:
I got this from one of the other files [here](https://github.com/spacetelescope/poppy/blob/c0840c52da1f87a5b2b8bbe264aa80c70b8550d9/poppy/instrument.py#L422-L425). This is because RTD parses the code without installing it, if it is not installed, it will not have a version file.

robelgeda:
Alright, I cleaned up all the test commits

PR #345, opened 2019-12-13
Adjust wavelength limit in instrument.py

ojustino:
I realized that, before #339, `nwavelengths` in instrument.py was probably limited to 100 because FITS header keys are only supposed to have up to 8 characters. In cases where `100 < nwavelengths <= 10000`, I introduce a different key style that can fit more digits.

This construction should be backward-compatible while also allowing up to 10,000 slices in data cubes going forward. I generated data cubes that fall into all three categories and the results matched my expectations.

### DIFF
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -310,7 +310,16 @@ class Instrument(object):
             you would supply in a call to calc_psf via the "monochromatic" option
         """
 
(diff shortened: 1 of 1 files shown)


mperrin:
As discussed in person - going to redirect this PR to `develop` rather than `master. Moved to #351

PR #347, opened 2020-01-23
Fix logspace() calls for numpy 1.18+

ehpor:
As of numpy 1.12 passing a non-integer number of samples for logspace() was deprecated. As of numpy 1.18, this is now mandatory, yielding an error when you do pass a float anyway.


mperrin:
Thanks @ehpor!

Travis test failed on Py3.5 because of an unrelated issue; that will be fixed in #348. OK to merge now despite the failing test.

Issue #297, opened 2019-02-25
Coveralls


robelgeda:
Coveralls is no longer reliable because github changed the way that 3rd party apps are integrated and coveralls appears to have done nothing to fix it. Its not clear if they will address this which means we should look for alternative services. One option that is currently popular is [Codecov](https://codecov.io). 

Related to #294 #296 
cc: @mperrin @shanosborne

mperrin:
I'm on board with switching to codecov. Is there recommended template setup for that we can lift from another ST package?

robelgeda:
I don't think ST has any templets for integration setup. I need to look into resources on how to.

mperrin:
@robelgeda have you been able to find any guidance for how to switch to codecov rather than coveralls? Do you think it's worth putting effort into this now, or leave it for some future time? Whichever you think best should be fine I think.

robelgeda:
I did not find anyone to help me with this. I think our best bet is to just try to do it ourselves. It would be a good learning experience as well I suppose. I am not sure if Coveralls is still not working so I will post on Slack to see what people say.

mperrin:
Codecov switch implemented in #349

PR #349, opened 2020-02-03
Remove astropy-helpers and implement codecov

### LINKED ISSUE #328: Remove use of astropy-helpers submodule
See astropy/astropy-APEs#52 ("A roadmap for package infrastructure without astropy-helpers") which describes how the functionality in astropy-helpers can be achieved via more standardized tools in the Python ecosystem. 

This is clearly good and we should do it. 

See for instance https://github.com/spacetelescope/mosviz/pull/179 and https://github.com/spacetelescope/cubeviz/pull/492 for examples of PRs that remove astropy-helpers. We should do similarly here.

shanosborne:
This PR attempts to remove the dependency on the `astropy-helpers` sub-package. The big changes are:
- All `setup.py` options have been moved to `setup.cfg` (with some over-arching requirements set in `pyproject.toml`
- All `travis` options have been moved to `tox.ini` which now controls testing, doc building, and coverage.
- Coverage is now run through `codecov` rather than `coveralls`
- `_astropy_init.py` has been removed and relevant information has been moved to `__init__.py`
- Updated docs structure so they would build correctly

Closes #328 and #297


mperrin:
I'll take a look - thanks!

One part I have been wondering about is the handling of version numbers via setuptools_scm. Do you understand how that part works, well enough to update our developer notes on making releases and so on?

.travis.yml
```diff
@@ -1,87 +1,50 @@
-language: python
-
-python:
-    - 3.7
+language: c
```
mperrin, line 1: This confuses me - why set the language to C when it's all Python code? 

(To be clear I'm not objecting if this is part of the recommended Travis setup. I'm just saying I don't understand it.)

tox.ini
```diff
@@ -0,0 +1,59 @@
...
+
+[testenv]
+passenv = *
+description = run tests with pytest
+deps =
+    pytest
+    legacy: numpy==1.13.*
+    legacy: astropy==3.2.*
+    astropydev: git+git://github.com/astropy/astropy
+    numexpr: numexpr>=2.0.0
```
mperrin, line 15: This part is really kind of slick. Nice.

.travis.yml
```diff
@@ -1,87 +1,50 @@
...
         # do the actual tests.
-        # Do a coverage test in Python 3.
-        - python: 3.7
-          env: SETUP_CMD='test --coverage'
+        # Do a test in Python 3.7
+        - env: TOXENV='coverage' TOX_ARGS=''
 
         # Check accelerated math version too
-        - python: 3.7
-          env: SETUP_CMD='test --coverage' PIP_DEPENDENCIES='numexpr'
```
mperrin, line 36: The existing version of this runs coverage tests on both the regular and with-numexpr versions. That results in showing higher coverage on the test report (which is a more accurate depiction of the actual test coverage, given that the numexpr test case exists). I assume codecov can probably also handle merging test reports from multiple test builds? 

(Not a priority for an immediate change, but perhaps you know enough of tox configs and codecov now to see how to add that in?)

mperrin (approved):
Much of this is beyond my detailed understanding of these tools and their necessary config files. So I'm placing a lot of weight in the fact that the tests all pass with this, which is pretty nice. 

A few comments / questions attached, but I think this seems good to go for merging into `develop`.

.travis.yml
pllim: @mperrin , this is "inherited" from the Astropy project. Python is not available natively on all OSes in Travis, so Astropy uses the `conda` version of Python via `ci-helpers`. Therefore, the "language" does not really matter as long as you grab the conda's Python via `ci-helpers`.

poppy/tests/coveragerc
```diff
@@ -1,16 +1,16 @@
...
+   poppy/_astropy_init*
+   poppy/conftest*
+   poppy/cython_version*
+   poppy/setup_package*
+   poppy/*/setup_package*
+   poppy/*/*/setup_package*
+   poppy/tests/*
+   poppy/*/tests/*
+   poppy/*/*/tests/*
+   poppy/version*
```
pllim, line 13: If you use `tox`, you also need to repeat these lines with `*/` in front, according to Tom R.

pllim: Also in the "final" guide published in http://docs.astropy.org/projects/package-template/en/latest/ape17.html#step-15-coverage-configuration , this file should be removed altogether and the config here moved to `setup.cfg`.

pllim (commented):
By removing the `_astropy_init.py` file, you removed the functionality of using Astropy test runner. Was that intentional?

mperrin:
@pllim - I thought the astropy test runner was deprecated in favor of just calling `tox` to run tests. Is that incorrect?

pllim:
They are two separate things, actually. `tox` is more for CI, while the test runner is for users to test a local installation. If you do not care about user being able to run something like `poppy.test()` from their Python session, then this PR is okay in that regard.

PR #350, opened 2020-02-04
Update to package structure

shanosborne:
A few left over changes from the previous PR #349. These include:
- Updating the travis setup to run coverage on the numexpr case as well as the "regular" case.
- Moving the contents of coveragerc to setup.cfg (as indicated in the new APE17 migration documentation here: http://docs.astropy.org/projects/package-template/en/latest/ape17.html#step-15-coverage-configuration)
- Removing reference to `poppy.test()` from documentation (we determined that this is a functionality we don't use and therefore want to keep it removed).


shanosborne:
The coverage commands aren't working (despite it working locally) - so this PR actually isn't passing right now.

shanosborne:
@mperrin this PR is ready to go for review - the coverage is now working for both runs (with and without numexpr)

setup.cfg
```diff
@@ -49,4 +49,63 @@ testpaths = poppy docs
...
+   # Don't complain if tests don't hit assertions
+   raise AssertionError
+   raise NotImplementedError
+
+   # Don't complain about script hooks
+   def main\(.*\):
+
+   # Ignore branches that don't pertain to this version of Python
+   pragma: py{ignore_python_version}
+omit =
```
mperrin, line 93: It seems annoying & suboptimal that this list of omits has to be repeated twice, but I'll take your word that's how it has to be set up...

tox.ini
```diff
@@ -1,14 +1,18 @@
 [tox]
 envlist =
     py{35,36,37}-test
-    py{35,36,37}-{astropydev,numexpr}-test
-    py35-legacy-test
+    py{35,36,37}-{astropydev,legacy}-test
+    py{35,36,37}-cov
+    py37-numexpr-cov
```
mperrin, line 6: Just for my own understanding - the above line says there's an environment called 'py37-numexpr-cov', but where does it actually define what happens inside that environment? It automatically infers what to do just based on there being an environment named that? That's neat.

tox.ini
```diff
@@ -36,19 +42,6 @@ conda_deps=
 commands=
     sphinx-build docs docs/_build
 
-[testenv:coverage]
```
mperrin, line 39: This deletes the `coverage` test environment? Just to check my understanding, after this the coverage tests are ran in the environments that it automatically configures based on the names above? So it doesn't need to have a specific environment created specifically. And uses the commands that are defined on the lines starting with "cov:". Yes?

mperrin (approved):
Looks good to me. Some questions just to double check my understanding of how the tox system works.

tox.ini
shanosborne: It's defined under `[testenv]`. So everywhere in the testenv where a line is preceded by `numexpr: `, `cov: `, or nothing (its the default, like pytest), the `py37-numexpr-cov` environment will pull that information. So it's all about the precursors in the deps and commands sections

tox.ini
shanosborne: Yes exactly. I moved the coverage info up to the main `[testenv]` section and everything starting with `cov:` is what it'll call for coverage runs.

setup.cfg
shanosborne: Yep I agree, but for some reason putting omit in the coverage run isn't enough - it'll still show up in the report.

PR #356, opened 2020-04-23
Update minimum supported Python version to 3.6

mperrin:
See also https://github.com/spacetelescope/webbpsf/pull/353


shanosborne (approved):
Looks good! We could increase the legacy versions of numpy and astropy now that it's a python 3.6 test, but since the versions still work together and pass travis I see no reason why it has to be changed.

PR #357, opened 2020-05-05
Allow Segmented_PTT_Basis to take in additional keyword arguments

kjbrooks:
I am using `Segmented_PTT_Basis` in the ST lab for creating commands to send to segmented deformable mirrors. We need to be able to specify custom apertures with this command and this can be done through the `segmentlist` variable in the `MultiHexagonAperture` class which is inherited by `HexSegmentedDeformableMirror` which is used in `Segmented_PTT_Basis`. However, to be able to access this from `Segmented_PTT_Basis`, we need to be able to pass in additional keyword arguments used by `HexSegmentedDeformableMirror` and therefore `MultiHexagonAperture`. 

This PR adds **kwargs to the`Segmented_PTT_Basis` input.

### DIFF
--- a/poppy/zernike.py
+++ b/poppy/zernike.py
@@ -820,7 +820,7 @@ def arbitrary_basis(aperture, nterms=15, rho=None, theta=None, outside=np.nan):
 
 class Segment_PTT_Basis(object):
     def __init__(self, rings=2, flattoflat=1*u.m, gap=1*u.cm, center=False,
-                pupil_diam=None):
+                pupil_diam=None, **kwargs):
         """
         Eigenbasis of segment pistons, tips, tilts.
         (Or of pistons only using the Segment_Piston_Basis subclass.)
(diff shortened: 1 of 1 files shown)


mperrin (approved):
Looks good! We can merge this as soon as the CI tests pass.

kjbrooks:
@mperrin is it unusual that checks haven't completed yet?

mperrin:
Huh. Yeah, that’s strange indeed. Must be something going wrong with Travis?

mperrin:
Huh. Yeah, that’s strange indeed. Must be something going wrong with Travis?

kjbrooks:
If you click on "Details" on the Pending build, it says that all the checks passed, so it seems like things are just not updating for some reason

kjbrooks:
I also received 7 emails notifying me of your comment, @mperrin so it seems likely that something is up with GitHub

PR #358, opened 2020-05-18
Release notes for 0.9.1, and add Travis automation of PyPI releases

mperrin:
Release notes. Now also includes PyPI automation for releases on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,4 +48,14 @@ after_success:
 
 notifications:
   email:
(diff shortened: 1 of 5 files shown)


shanosborne (approved):
This looks good to me. The travis.yml update matches what's in pysiaf except for the username line, and in that case what's here is what matches the travis docs so I assume it should work. And the release guide looks good - its very clear

PR #368, opened 2020-06-22
Update utils.py so that measure_FWHM works on integer as well as float dtype

### LINKED ISSUE #363: Potential bug in poppy.measure_fwhm
In attempting to use poppy.utils.measure_fwhm, I ran into a TypeError: No loop matching the specified signature and casting was found for ufunc true_divide. 

It seems that this line: 

https://github.com/spacetelescope/poppy/blob/2010aaac2e738ac347186e28b2258e489a22deec/poppy/utils.py#L821

is causing an error since this operation is not actually valid for the reasons laid out in this issue: https://github.com/numpy/numpy/issues/10565/

It seems like a relatively simple solution of change that line to: `image = image / image.max()`

kjbrooks:
If the input fits image is integer dtype, the current code breaks. Making this change will make this work for both float and int arrays.

Closes #363

### DIFF
--- a/poppy/utils.py
+++ b/poppy/utils.py
@@ -829,7 +829,7 @@ def measure_fwhm(HDUlist_or_filename, ext=0, center=None, plot=False, threshold=
         raise ValueError("input must be a filename or HDUlist")
 
     image = HDUlist[ext].data.copy()  # don't change normalization of actual input array; work with a copy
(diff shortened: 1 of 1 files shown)


mperrin (approved):
Looks good!

@kjbrooks this is optional but do you think we should add a test case to verify that the fix works and the function operates correctly on both float and int types?

PR #371, opened 2020-06-25
fix issue with rotation of HexSegDMs

mperrin:
Fix coordinated issue with influence functions for rotation of hexagonally segmented DMs. Adds unit test for the correct behavior too.


kjbrooks (approved):
Looks good to me!

PR #376, opened 2020-08-19
adding test function for sub sampled optics/shwfs, fixing bug in call to SquareAperture

remorgan123:
Hi @mperrin and @douglase , here is a first pass at adding an automatic test for the sub sampled optics class. It creates a Shack Hartmann Wavefront Sensor object and uses it to sample a flat wavefront and a wavefront after bouncing off an actuated deformable mirror and asserts that the resultant reconstructed wavefront is nonzero. I tried to base this formatting off of the other files in the tests folder, let me know if it needs to be changed. 

Other small fixes in this PR: 
- there was an issue in the call to square aperture when initializing a SHWFS with square lenslets which is fixed here
- I added a note before the call to fwcentroid to suggest what to do if fwcentroid gives an index error. Please let me know if you have anything to add to that note. 

Please let me know if you have any suggestions for additional tests that would be useful to add here! Thanks!


mperrin:
Hi @remorgan123 and happy new year! I'm (finally, belatedly) getting around to looking at this PR. I just sent an update to your branch to resolve a conflict with the current poppy develop, and will check now that the test runs on this. Thanks again.

mperrin (approved):
Looks good to me! I made one change to the test function: slimming it down from a 12x12 Shack Hartmann to just a 2x2 for the test case, which still tests all the functionality but reduces the time it takes the test to run from about 75 s to 2 s.  So this is a more efficient  test approach in terms of runtime. 

(To be clear I have no objection to longer runtime of a test if it adds value; but in this case the "test the SH" goal is not dependent on any particular size of array).

Thanks again @remorgan123

remorgan123:
Sounds good, thanks for the review and the update!

PR #377, opened 2020-08-27
Use separate numpy.random.RandomState in statistical PSD WFE

ivalaginja:
The class `StatisticalPSDWFE` sets a seed in the global numpy random number generator. Since it does this for the global RNG, every other random proccess that uses it will have the same seed set, and prevents random number generation elsewhere. This is easily solvable by making instances of `numpy.random.RandomState` in each separate random process, which also means that the seed can be set separately.

This PR only introduces a separate random state for `StatisticalPSDWFE` in `wfe.py`, however, this could be done in any place where `numpy.random` is used, for more clarity and robustness.

### DIFF
--- a/poppy/wfe.py
+++ b/poppy/wfe.py
@@ -345,8 +345,9 @@ class StatisticalPSDWFE(WavefrontError):
         rho, theta = _wave_y_x_to_rho_theta(y, x, self.radius.to(u.meter).value)
         psd = np.power(rho, -self.index)   # generate power-law PSD
 
-        np.random.seed(self.seed)   # if provided, set a seed for random number generator
-        rndm_phase = np.random.normal(size=(len(y), len(x)))   # generate random phase screen
+        psd_random_state = np.random.RandomState()
+        psd_random_state.seed(self.seed)   # if provided, set a seed for random number generator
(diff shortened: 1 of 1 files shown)


ivalaginja:
As pointed out by @mperrin:
> I’ve noticed that the Numpy docs refer to RandomStare as “legacy random number generation”: https://numpy.org/doc/stable/reference/random/legacy.html
> 
> The current recommended method appears to be Generator instead: https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator
> 
> I won’t pretend to know anything technical about the difference between those, nor if RandomState might be deprecated in some future numpy release. For our purposes probably either is ok (and clearly better than now). But I wonder if we should prefer to use Generator now.

That is correct, and I can change this to use `numpy.random.Generator` instead. This would elevate the numpy requirement to >1.16.0 though, is it fine if I go ahead and change that in the `environment.yml` too?

mperrin (approved):
Good fix. This prevents the setting of the random number seed for this WFE optic from leaking out to affect other random number uses in other code, in ways that may not be obvious. (And were not obvious, when we ran into this issue on HICAT)

Issue #373, opened 2020-08-03
Integration w/astropy units, models?


keflavich:
I'd like to use `poppy` to make synthetic PSFs to use in modeling with `astropy.models`, so it would be nice if they played well together.  Right now, afaict, poppy is not `unit`-friendly - is refactoring to include units in the works?  Notably, some of the examples specify units ambiguously (e.g., https://github.com/spacetelescope/poppy/blame/develop/docs/examples.rst#L32 says "microns" but the value is in meters).

Are there any plans to incorporate units and/or modeling into poppy?

mperrin:
Hi @keflavich  - in fact Poppy *is* units-friendly, and makes extensive usage of astropy.units throughout the code, though some of the docs are out of date and don't reflect that clearly. (Since some of the docs were well written before `astropy` itself even existed.)

Try the same function call with wavelength equal to `2e-6`, `2*u.micron`, `2e-6*u.meter` and you will find the same results in all cases. Numbers without units are interpreted as meters, but it's totally fine and supported to supply inputs as quantities with units too. 

This applies to many/most of the optical element classes as well: `CircularAperture(radius=300*u.centimeter)` is equivalent to `CircularAperture(radius=3*u.meter)` is equivalent to `CircularAperture(radius=3)`. Image plane elements can be specified in `u.arcsec` or other angular units, detector pixel scales can be specified as `0.032*u.arcsec/u.pixel`, and so on. 

What particular functionality are you looking for with units and with modeling? FWIW we have already worked to make sure that output PSFs can be generated consistent with the `photutils` ePSF modeling framework; see the gridded library code in WebbPSF.

That said, the output PSF data arrays do not have units associated with them (i.e. the `psf`  HDUList `.data` attributes are just plain ndarrays, not Quantities). At one point I benchmarked doing the optical calculations directly on Quantity arrays but found a significant slowdown compared to calculations on pure ndarrays. So for performance reasons internally in the propagation calculation code, Poppy casts quantities to standard units (meters, arcsec) and pulls out the bare values for use in the FFT/MFT calculations.

keflavich:
Huh, thanks.  I think I had put in a wrong unit, then, and misinterpreted the error message as 'units aren't supported' - I put in arcsec for pixel scale, instead of arcsec/pixel.  

[...]

PR #378, opened 2020-09-10
Improve some uses of astropy units

mperrin:
Addresses comments on #373 by @mcbeth

- [x] fix use of units in computation of average wavelength FITS header keyword. The FITS header keyword WAVELEN is now always in units of meters, correctly, regardless of whether some other unit is specified as part of the PSF computation.  

- [x] check/fix use of units in detector pixel scales, to fix problems when setting detector pixel scale to something not in arcseconds, e.g. `100*u.nrad/u.pixel`. It turns out that the PSF computation itself was handling the units correctly - but the display functions on the other hand were not appropriately handling the different pixel scale units. I implemented a fix for that.  Further, I then made  related improvements to the display functions `Wavefront.display()` and `utils.display_PSF()`, to allow specifying PSF or wavefront display using any desired angular unit, not just arc seconds, for the X and Y axes scales.

As a result you can now do for instance the following, and all the unit handling works as expected: 

```python

osys = poppy.OpticalSystem()
circular_aperture = poppy.CircularAperture(radius=1*u.m)
osys.add_pupil(circular_aperture)
zernike_coefficients_sequence = [0, 100e-9, 0, -50e-9, 0, 35e-9]. #semi-arbitrary WFE
thinlens = poppy.ZernikeWFE(radius=1*u.m, coefficients=zernike_coefficients_sequence)
osys.add_pupil(thinlens)
osys.add_detector(pixelscale=0.02*u.urad/u.pixel, fov_pixels=128)

psf_with_zernikewfe = osys.calc_psf(wavelength=400*u.nm, display_intermediates=True)

plt.figure()
poppy.display_psf(psf_with_zernikewfe, angular_coordinate_unit=u.urad, markcentroid=True)
```
Unknown-1
Unknown

(None of this directly helps any of the immediate use cases in webbpsf, but it does help make `poppy` more robust and extensible through more consistent use of the astropy.units framework)

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -333,7 +333,8 @@ class BaseWavefront(ABC):
     def display(self, what='intensity', nrows=1, row=1, showpadding=False,
                 imagecrop=None, pupilcrop=None,
                 colorbar=False, crosshairs=False, ax=None, title=None, vmin=None,
-                vmax=None, scale=None, use_angular_coordinates=None):
+                vmax=None, scale=None, use_angular_coordinates=None,
+                angular_coordinate_unit=u.arcsec):
         """Display wavefront on screen
 
         Parameters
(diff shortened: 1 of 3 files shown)


mperrin:
Hmm, here's another issue with units that should work but doesn't:
```
ap1 = poppy.RectangleAperture(width=50*u.micron, height=300*u.micron, shift_x = 100*u.micron)
ap1.display()
```
You can't specify shift_x or shift_y with units, just as a bare float with implicit units. This too should be cleaned up and made consistent & unsurprising.

shanosborne (approved):
The only thing I see is that we should probably document somewhere that the units need to be fixed in the `poppy.RectangleAperture` case that you mentioned in a comment, since this change isn’t included in this PR. Other than that, this all looks good to me!

PR #379, opened 2020-09-14
Fix warning on matplotlib cmap modifications in place

spacegal-spiff:
Multiple deprecated matplotlib warnings were occurring when calling the display function in poppy_core.py:
```
/Users/kwduong/miniconda3/envs/pandeia/lib/python3.7/site-packages/poppy/poppy_core.py:2459: 

MatplotlibDeprecationWarning: You are modifying the state of a globally registered colormap. 

In future versions, you will not be able to modify a registered colormap in-place. 

To remove this warning, you can make a copy of the colormap first. cmap = copy.copy(mpl.cm.get_cmap("RdBu_r"))
  cmap_opd.set_bad('0.3’)
```
I fixed these throughout with the suggested use of copy.copy() when getting colormap attributes from the conf file.

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -457,15 +457,15 @@ class BaseWavefront(ABC):
             if vmin is None:
                 vmin = 0.25 if what == 'phase' else 0
             norm_inten = matplotlib.colors.Normalize(vmin=vmin, vmax=vmax)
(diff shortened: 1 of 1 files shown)


mperrin (approved):
Looks good. This resolves a new warning that was introduced recently in matplotlib 3.3.0 (see https://github.com/matplotlib/matplotlib/pull/16991), and is needed prior to matplotlib 3.5.

PR #380, opened 2020-09-16
Improve radial_profile to allow measurement of partial profiles for sources offset outside the FOV

mperrin:
In some cases it is useful to be able to compute a partial radial profile for a source that is offset outside out of the field of view. (Specific motivating use case: the JWST NIRSpec IFU and/or MIRI MRS, hypothetically observing a faint companion near a bright star, with the bright star just outside the IFU field of view). 

This PR makes a small change in the radial_profile function to correctly compute the returned radius coordinates in this case. This fixes a bug in the current code that always started the radius array at zero. The user still has to pass in the correct center location for the PSF using the existing `center` keyword, same as for a PSF offset within the FOV. 

A unit test demonstrates consistency of the profile for a centered and offset source in the overlap region. The profiles are checked for consistency at each of the test radii indicated by the vertical lines. 

Unknown-7

### DIFF
--- a/poppy/tests/test_utils.py
+++ b/poppy/tests/test_utils.py
@@ -131,6 +131,56 @@ def test_radial_profile(plot=False):
     assert np.allclose(prof2, prof3)
     # TODO compare those to be near a sinc profile as expected?
 
+def test_radial_profile_of_offset_source():
+    """Test that we can compute the radial profile for a source slightly outside the FOV,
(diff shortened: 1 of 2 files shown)


mperrin:
FYI @k2wilcomb

shanosborne (approved):
This looks good to me. I would suggest just re-starting the travis job and let it re-run. It errored because of an HTTP error and those can usually be fixed just be re-running the code.

PR #383, opened 2020-10-29
Power Spectrum WFE

jlumbres:
Hello POPPY team,

Here's code Jared Males (@jaredmales) and I have been developing for generating a comprehensive power spectrum WFE. We calculate the PSD based on the power law with outer scale, inner scale, and surface roughness incorporated. Current documentation exists so far on the MagAO-X PDR Section 5.1 (https://magao-x.org/docs/handbook/appendices/pdr/). A publication with this work applied to MagAO-X optical surfaces is in progress.

This code assumes an optical surface's power spectrum has been characterized using the modeling parameters. It won't go into how to get those parameter values from a measured power spectrum (that is coming in the paper), but it will show how using those parameters together make the WFE opd map. The code structure is based on Iva Laginja's (@ivalaginja) StatisticalPSDWFE class.

I have a demo notebook posted in "notebooks/" that gives a visualization of the process and how to use basic features in PowerSpectrumWFE. Some features of PowerSpectrumWFE include:
- establishing desired wavefront error rms
- applying wfe scaling for reflective surfaces
- setting the PSD screen size
- setting a seed value to consistently produce the same randomized surface
- capable of building a wfe opd with multiple PSD parameter sets incorporated

I included a test case posted in “tests/test_wfe.py” (with generous assistance from Joseph Long (@joseph-long) and Ewan Douglas (@douglase)). Since there's no pre-existing functions for comparison testing, the best we could do is a test of self-consistency. Jared independently generated an opd surface with specific optical and PSD parameters. My test shows that using the same set of parameters, I was able to generate a comparable surface with the same RMS. There is a small problem in that the test case requires reading in a .fits file. We're not sure how to proceed without it, so we welcome discussion on it.

Please let me know if you have any questions about the code, suggestions for improving the demo / test / documentation, and anything needed for implementing it with POPPY.

Thanks,
Jennifer


mperrin:
Thanks @jlumbres for this contribution! I will try to take a look at this, but can't promise to do so within the next couple of weeks given various deadlines. Much appreciated in any case. 

In terms of adding the test data file, I see it's only 2 MB, which is not much; I don't have any particular objection to including that as a data file within this package.

poppy/wfe.py
```diff
@@ -353,6 +353,151 @@ def get_opd(self, wave):
...
+        self.opd = phase_screen / np.std(phase_screen) * self.wfe.to(u.m).value  # normalize to wanted input rms wfe
 
-        return opd
+        return self.opd
+
+
+class PowerSpectrumWFE(WavefrontError):
+    """
+    Power spectrum PSD WFE class from characterizing and modeling 
+    optical surface power spectrum and applying optical noise.
```
mperrin: We should spell out here the acronym, like: "power spectral density (PSD)."  
The phrasing here could also be clarified slightly - I think the current wording is a bit awkward. Perhaps more like: 
```
WFE model specified via a Power Spectral Density (PSD),
or a list of multiple PSDs
```

I'd also encourage including the equations here as LaTeX, which will get correctly rendered into the online docs.

poppy/wfe.py
```diff
@@ -353,6 +353,151 @@ def get_opd(self, wave):
...
+    Lumbres, et al. In Prep.
+
+    Parameters
+    ----------
+    name : string
+        name of the optic
+    psd_parameters: list of array with various astropy quantities
+        Specifies the various PSD parameters with appropriate units, 
+        ordered according to which PSD model set. Follows the order:
+        [alpha, beta, outer_scale, inner_scale, surf_roughness]
```
mperrin: This part isn't entirely clear to me. What does "ordered according to which PSD model set" mean? Also, what are appropriate units for each of these quantities? Lastly, the description of the argument type says it's a list of arrays, so is each alpha itself a numpy ndarray? Or do you mean it is a list of quantities? 

Moreover, what are the meanings of each of alpha, beta, outer_scale, inner_scale, surf_roughness? These should be stated here in the doc string, perhaps just by quoting appropriate definitions from the reference in the MagAO docs.

poppy/wfe.py
```diff
@@ -353,6 +353,151 @@ def get_opd(self, wave):
...
+        Seed for the random phase screen generator
+    apply_reflection: boolean
+        Applies 1/2 scale for the OPD as needed for reflection
+        Set to True if the PSD model does not account already for reflection.
+    screen_size: integer
+        Sets how large the PSD matrix will be calculated.
+        If None passed in, then code will default size to 4x wavefront's side.
+    wfe: astropy quantity
+        Optional. Use this to force the wfe RMS for opd surface.
+        If None passed, then the wfe RMS produced is what shows up in PSD calculation.
```
mperrin: perhaps this parameter should be called `rms` or `normalization`? Just `wfe` is ambiguous.

poppy/wfe.py
```diff
@@ -353,6 +353,151 @@ def get_opd(self, wave):
...
+        # set the random noise
+        psd_random = np.random.RandomState()
+        psd_random.seed(self.seed)
+        rndm_noise = np.fft.fftshift(np.fft.fft2(psd_random.normal(size=(self.screen_size, self.screen_size))))
+        
+        psd_scaled = (np.sqrt(psd/(pixelscale.value**2)) * rndm_noise)
+        opd = ((np.fft.ifft2(np.fft.ifftshift(psd_scaled)).real*surf_unit).to(u.m)).value 
+        
+        if self.apply_reflection == True:
+            opd = opd/2
```
mperrin: This seems precisely flipped to me: For a reflection, the wavefront error should be twice the surface error.  See for instance the `include_factor_of_two` parameter to the deformable mirror class. Or am I missing something different in this context?

poppy/wfe.py
```diff
@@ -353,6 +353,151 @@ def get_opd(self, wave):
...
+        super().__init__(name=name, **kwargs)
+        self.psd_parameters = psd_parameters
+        self.seed = seed
+        self.apply_reflection = apply_reflection
+        self.screen_size = screen_size
+        
+        if hasattr(wfe, 'unit') or wfe is None:
+            self.wfe = wfe
+        else:
+            self.wfe = wfe * u.m # default assume meters
```
mperrin: This part should be unnecessary, given the `utils.quantity_input` decorator. That has the effect of ensuring that the supplied value of that parameter has the right unit (or an equivalent convertible one), or adding that unit if it's given a bare float. In this case, if the user supplies a float it will automatically get treated as nanometers and converted into `u.nm`, before even getting to this if/else statement.

poppy/wfe.py
```diff
@@ -353,6 +353,151 @@ def get_opd(self, wave):
...
+    psd_weight: iterable list of floats
+        Specifies the weight muliplier to set onto each model PSD
+    seed : integer
+        Seed for the random phase screen generator
+    apply_reflection: boolean
+        Applies 1/2 scale for the OPD as needed for reflection
+        Set to True if the PSD model does not account already for reflection.
+    screen_size: integer
+        Sets how large the PSD matrix will be calculated.
+        If None passed in, then code will default size to 4x wavefront's side.
```
mperrin, line 429: Consider adding here some brief explanation for why the user might want to adjust this, and what effect it has. It's probably not obvious to a non-expert why it's desirable to make this much larger than the wavefront.

poppy/wfe.py
```diff
@@ -353,6 +353,151 @@ def get_opd(self, wave):
...
+        
+        if self.apply_reflection == True:
+            opd = opd/2
+            
+        if self.screen_size > wave.shape[0]: # crop to wave shape
+            opd = utils.pad_or_crop_to_shape(array=opd, target_shape=wave.shape)
+            
+        if self.wfe is not None:
+            rms = np.sqrt(np.mean(np.square(opd)))
+            opd = opd * (self.wfe.to(u.m).value/rms)
```
mperrin: Minor, but this can be rewritten as `opd *= self.wfe.to(u.m).value/rms`, which is equivalent but slightly more performant.

mperrin (changes requested):
Hi @jlumbres and happy new year. I've (finally, belatedly) gotten around to reading and reviewing your power spectrum PR. This looks good overall! I do have some suggestions and small requests for some clarifications in the documentation, but the code itself looks good. 

Thanks again for contributing this. Cheers.

mperrin:
Hi @jlumbres, I wanted to check in about this PR since I hadn't heard back yet from you about the review comments I sent a while ago. It's fine and understandable if you've moved on to other efforts lately - I just wanted to check if you had a timescale in mind for returning to this, or if we should set it aside. I'm fine with things either way - I ask just because I'm trying to organize the next release of poppy v1.0 and I'd like to know if we should try to include this or not. Cheers and thanks.

jlumbres:
Hi @mperrin - thanks for checking in and sorry for falling off on things. I've been spending a lot of time writing my paper on modeling which uses this code. It looks like most of your requested changes are documentation related and a lot of it are stuff I've been writing for the paper to some extent. How soon does your timeline look for adding this to v1.0? I'm close to hitting a stable point where I can spend time on this, but I'd like to not be your bottleneck.

mperrin:
Hi @jlumbres - I had been hoping to wrap up the poppy 1.0 release this month (March) but clearly enough didn't get there yet.  Hopefully next week maybe. It sounds like it may be best to leave this PR aside for now, and we can include it in 1.1 sometime soon enough. I do think this is very useful and it'll be great to include, but we're all juggling lots of competing demands on our time and I totally get that your paper takes priority right now (as it should!).

poppy/wfe.py
jlumbres: No, you were correct here. I was going backwards in my head. I have corrected it to be x2 instead of /2 in the update.

poppy/wfe.py
jlumbres: Okay. I have removed this block.

poppy/wfe.py
jlumbres: Okay. I have edited this in the comment block. I think my reasoning is sufficient? It could use more detail, maybe.

poppy/wfe.py
jlumbres: Oh no you're seeing my rudimentary python coding. I've fixed this through my code. Thanks for the tip.

poppy/wfe.py
jlumbres: I have changed the parameter to rms, as suggested.

poppy/wfe.py
jlumbres: I've added more content about the von Karman PSD model in the comment block. However, I'm not quite sure how to set up the LaTeX for the comments. Joseph pointed me to something I can use, but I'm unsure if only because I don't know how to check my writeup (like with all LaTeX writing practice). If you can let me know how I can check this, I can fix it up.

poppy/wfe.py
```diff
@@ -353,6 +353,231 @@ def get_opd(self, wave):
...
-        return opd
+        return self.opd
+
+
+class PowerSpectrumWFE(WavefrontError):
+    """
+    WFE model specificed via a Power Spectral Density (PSD),
+    or a list of multiple PSDs, which follow von Karman PSD model:
+    
+    :math: `P(k) = \frac{\beta} {\left( \left(\frac{1}{L_{0}}\right)^{2} + |k|^{2} \right)^{{\alpha/2}}} e^{-(|k|l_{0})^{2}} + \beta_{sr}`
```
joseph-long: You’ll want to remove the space between the directive and the text enclosed in backticks for it to render properly. Since backticks are used for formatting I couldn’t give an explicit example in Slack from my phone 😛here’s a better one

```
:math:`x^2`
```

poppy/wfe.py
jlumbres: This one is slightly tricky. We're taking a piecewise approach since a single set of parameters doesn't quite fit well with modeling the measured PSD. This is supposed to be a list of lists, where each list element is a list of the different PSD model parameter quantities. So, if a PSD needs to be modeled by combining 2 PSDs together, then it would show up as` [ [alpha1, beta1, outer_scale1, inner_scale1, surf_roughness1], [alpha2, beta2, outer_scale2, inner_scale2, surf_roughness2]]`
Does this make sense? I'm open for discussing this better so the documentation makes sense.

poppy/wfe.py
```diff
@@ -353,6 +353,231 @@ def get_opd(self, wave):
...
         phase_screen -= np.mean(phase_screen)  # force zero-mean
-        opd = phase_screen / np.std(phase_screen) * self.wfe.to(u.m).value  # normalize to wanted input rms wfe
+        self.opd = phase_screen / np.std(phase_screen) * self.wfe.to(u.m).value  # normalize to wanted input rms wfe
 
-        return opd
+        return self.opd
+
+
+class PowerSpectrumWFE(WavefrontError):
+    """
```
joseph-long: For boring technical reasons you will want this to be a “raw string”. Just prefix the quotation marks with an r:

```
r“””foo”””
```

jlumbres: Is it for the entire raw string only? Considering how large the comment block is.

poppy/wfe.py
jlumbres: Okay, thanks for the correction. Does this continue to hold if I wanted to have a string next to the LaTeX variable? For example: 
:math:`L_{0}`: astropy quantity
edit: I've given up finding the correct markup option, sorry if it doesn't make sense

poppy/wfe.py
joseph-long: Yes, putting that r marks the whole docstring as raw. If you don’t mark it as a raw string, the backslashes in latex commands are interpreted as character escape sequences... e.g. \nabla being “newline character followed by ‘abla’”

if you’re using character escapes and latex syntax in the same string, of course, heaven help you. But that shouldn’t arise in a docstring.

poppy/wfe.py
joseph-long: Yep backticks are overloaded. Formatting formatting syntax is hard.

joseph-long: The parser sees it as

```
:<name>:`<content>`
```

So anything after the second backtick shouldn’t matter. You could put a colon right after, or a comma, or another colon thing (i think docutils calls them “roles”?)

jlumbres:
Hi @mperrin - I've hit a stable point in writing my paper (that is, I finished writing the paper and it's going through the comment cycle), so I've made the documentation updates. 

I included an additional parameter `incident_angle`, which will scale the rms based on surface beam distortion from reflection. Since the surface rms is based on paraxial measurements, the `incident_angle` will scale the rms in the code to match reflected beam distortion rms. It doesn't change the beam in POPPY to be distorted (still remains a circular beam), but it will now contain the distorted rms value. I did this to match the MagAO-X instrument model as closely possible.

I'll fix the test code and notebook demo later, once I can confirm that the code is working.

It also looks like the Travis CI build is failing, and I'm not sure how much of it is me or something further in the back end.

poppy/wfe.py
mperrin: No worries! We're all learning as we go. :-)

(I have spent a probably-disproportionate amount of time pursuing small gains in runtime throughout poppy, for lots of years now, so all these little optimizations are likely more apparent to me than they would be to most people)

poppy/wfe.py
```diff
@@ -353,6 +353,231 @@ def get_opd(self, wave):
...
+            self.screen_size = wave.shape[0]
+            
+            if wave.ispadded is False: # sometimes the wave is not padded.
+                self.screen_size = self.screen_size * 4 # default 4x, open for discussion
+        
+        elif self.screen_size < wave_size:
+            raise Exception('PSD screen size smaller than wavefront size, recommend at least 2x larger')
+        
+        # get pixelscale to calculate spatial frequency spacing
+        pixelscale =  * u.pixel # default setting is m/pix, force to meter only
```
mperrin: FYI - This looks to be the offending line that's making the tests fail. Looks like there's a missing variable between the = and * symbols.

poppy/wfe.py
```diff
@@ -353,6 +353,231 @@ def get_opd(self, wave):
...
+        Section 5.1: Optics Specifications, Eqn 1
+        https://magao-x.org/docs/handbook/appendices/pdr/
+    Lumbres, et al. In Prep.
 
-        return opd
+    Parameters
+    ----------
+    name : string
+        name of the optic
+    psd_parameters: list of lists
```
mperrin: `psd_parameters: list of lists`

Minor clarification.  Is this actually `psd_parameters: list (for single PSD), or list of lists (for multiple PSDs)`? I think so, from what you've written below, but want to check my understanding.

poppy/wfe.py
```diff
@@ -353,6 +353,231 @@ def get_opd(self, wave):
...
+        The PSD matrix needs to be larger than the wavefront for Fourier transform padding purposes.
+        If None passed in, then code will default size to 4x wavefront's side.
+        Default to None.
+    rms: astropy quantity
+        Optional. Use this to force the wfe RMS
+        If a value is passed in, this is the surface rms value (not OPD) in meters.
+        If None passed, then the wfe RMS produced is what shows up in PSD calculation.
+        Default to None.
+    incident_angle: astropy quantity
+        Adjusts the WFE based on
```
mperrin: Doc sentence cuts off halfway.

poppy/wfe.py
```diff
@@ -353,6 +353,231 @@ def get_opd(self, wave):
...
+        if hasattr(incident_angle, 'unit'):
+            if incident_angle.unit == u.deg:
+                assert incident_angle.value < 90, "Incident angle must be less than 90 degrees."
+            elif incident_angle.unit == u.rad:
+                assert incident_angle.value < (np.pi/2), "Incident angle must be less than pi/2 radians."
+            else:
+                raise Exception('Incident angle units must be either degrees or radians.')
+            self.incident_angle = incident_angle
+        else:
+            raise Exception('Incident angle missing units, must be either degrees or radians.')
```
mperrin: This piece can be automated by adding it to the quantity input wrapper above the function. Add incident angle to that, like `@utils.quantity_input(wfe=u.nm, radius=u.meter, incident_angle=u.deg)`. It will automatically enforce that any input to incident angle has to be in degrees or some other equivalent convertible unit (radians, arcmin, anything angular). 

That won't apply the range check relative to 90º, but that can also be simplified using units too, like this: 
```
if incident_angle >= 90*u.deg: 
    raise ValueError("Incident angle must be less than 90 degrees, or equivalent in other units")
```
That relies on the units framework to handle the conversions. If incident angle is given in radians or any other angular unit, when it's compared to a value in degrees the conversion will happen automatically.

poppy/wfe.py
jlumbres: That line was supposed to get deleted when I changed everything to be wave.pixscale. Let me submit the correction.

mperrin (commented):
Looks good. A few minor, optional comments below. I think this'll be good to approve and merge once we can get the tests passing. Thanks @jlumbres! (and also to @joseph-long for help with latex wrangling)

jlumbres:
Hmmmm, still failing. I think there may be something else I changed without noticing. I'm AFK for the majority of today, so I'll check up on this tonight/this weekend/early next week. Thanks @mperrin and @joseph-long for looking thoroughly through this with me.

mperrin:
Hi @jlumbres, I think the next fix is also straightforward: you'll need to update the test function to use `rms` instead of `wfe` for the name of that parameter.  

If you're not familiar with the testing system, you can click through the "Details" link to Travis, then click any of the test cases with a red X to see the test logs and error messages. Usually but not always this makes it relatively easy to figure out what's going on.

I'll look out for the update to this by early next week or so. Cheers.

poppy/wfe.py
jlumbres: Yes, this is correct.

poppy/wfe.py
jlumbres: Wow that's cool! Once again you're finding more of my rudimentary python coding. I made the fix, thanks for the tip.

poppy/wfe.py
jlumbres: Thanks for pointing it out. I finished the sentence. (I shouldn't be allowed to write docs at 11 PM)

jlumbres:
Hi @mperrin, I've made the fixes from your comments and also made other fixes in the test function. I'd like to thank @douglase for helping me offline and for reminding me how to test locally. I found a few little things that I didn't track well in the main code (`dk` was set to `.value` too soon, for instance). Looks like the Travis CI build passed, too. I think things look done for now, let me know if I need to make any additional tweaks to the documentation.

poppy/wfe.py
```diff
@@ -353,6 +353,225 @@ def get_opd(self, wave):
...
+            raise Exception('PSD screen size smaller than wavefront size, recommend at least 2x larger')
+        
+        # get pixelscale to calculate spatial frequency spacing
+        dk = 1/(self.screen_size * wave.pixelscale * u.pix) # eliminate the pixel units
+        
+        # build spatial frequency map
+        cen = int(self.screen_size/2)
+        maskY, maskX = np.mgrid[-cen:cen, -cen:cen]
+        ky = maskY*dk.value
+        kx = maskX*dk.value
```
mperrin: Minor: These will get the value of `dk` but possibly in unexpected units; if the input wavefront units are in anything other than meters/pixel, this won't cast to meters automatically as is. Here I suggest `dk.to_value(1./u.m)` to be sure you get out something that is actually 1/m as the next line says.

poppy/wfe.py
```diff
@@ -353,6 +353,225 @@ def get_opd(self, wave):
...
+        
+        # Set rms value based on the active region of beam
+        if self.rms is not None:
+            beam_diam = wave.pixelscale * u.pix * wave_size # has units, needed for circ
+            circ = CircularAperture(name='beam diameter', radius=beam_diam/2)
+            ap = circ.get_transmission(wave)
+            opd_crop = utils.pad_or_crop_to_shape(array=opd, target_shape=wave.shape)
+            active_ap = opd_crop[ap==True]
+            rms_measure = np.sqrt(np.mean(np.square(active_ap))) # measured rms from aperture
+            opd *= self.rms.to(u.m).value/rms_measure # appropriately scales entire OPD
```
mperrin: Couple questions about this part. This makes the area used in computing the RMS of this optic depend, not on this optic itself, but on the incident beam it is illuminated with. It also implicitly assumes that the illuminated beam diameter is the same as the full input wavefront (this may not be the case, if the wavefront has been stopped down by some prior aperture). 

The way this is handled for the ZernikeWFE or ParameterizedWFE classes is that they have a `radius` attribute on the WFE optic itself, which in effect defines the clear aperture radius of that optic, over which the WFE is to be evaluated. I wonder if that approach should also be used here. One benefit would be to make more consistent the API between all three of these classes, so they each have a radius parameter consistently. What do you think?

mperrin (approved):
Thanks @jlumbres this looks very good. I think we could proceed with this as is, but I have two small further suggestions/questions; See below. 

If you want, I can take over and add the small revisions for both of those, or you can yourself if you agree. In any case I think this is definitely getting to the finish line. I really do appreciate your efforts on this - it takes some extra work making things broadly usable for other people, and well-documented and robust, but I think it's worth it in the long run. Thanks again.

poppy/wfe.py
jlumbres: Okay, I have made the edit. Thank you for the suggestion.

poppy/wfe.py
jlumbres: The radius parameter makes a lot of sense. I've been so accustomed to the MagAO-X optics being consistent with their optical diameters across the board that I never thought of it that way. I also in favor of maintaining consistency, especially if it helps future users. I looked at the other classes and it's easy to implement. I'll include it with my next commit shortly. It might take a bit because I'll also need to fix the test case at the same time.

jlumbres:
Hi @mperrin - the edits are in and passed through Travis CI. I made an additional commit since then because I forgot the `radius` parameter documentation.

I appreciate your comments and suggestions through the process. I learned a lot writing python making this class. It's definitely a completely different experience to write code that you originally only wrote for yourself/very few people and then make edits so it it's broadly usable. Thanks again for the opportunity to add my work into the package and helping me make it possible.

Issue #40, opened 2018-08-23
`offset` argument to Detector class doesn't actually work


mperrin:
<a href="https://github.com/mperrin"><img src="https://avatars0.githubusercontent.com/u/1151745?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mperrin](https://github.com/mperrin)**
_Friday Jan 23, 2015 at 06:00 GMT_
_Originally opened as https://github.com/mperrin/poppy/issues/40_

----

Looks like I left a few minor bits out of this implementation. Will fix. 

Somewhat more complicatedly, _propagateMFTinverse doesn't support offset image planes. 

I actually no longer remember the details of what motivated this feature - but I just came across a use case in making one of the MFT to FFT comparison tests.   (Detector always uses ADJUSTABLE style MFTs, so there's no way to get exact equivalency with FFT outputs -- unless one uses this `offset` argument!)

mperrin:
<a href="https://github.com/mperrin"><img src="https://avatars0.githubusercontent.com/u/1151745?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mperrin](https://github.com/mperrin)**
_Tuesday Nov 17, 2015 at 16:53 GMT_

----

not likely to actually have time to look at this before 0.4, alas.

mperrin:
<a href="https://github.com/mperrin"><img src="https://avatars0.githubusercontent.com/u/1151745?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mperrin](https://github.com/mperrin)**
_Monday Sep 26, 2016 at 17:30 GMT_

----

nobody's asked about this, not a high priority.

mperrin:
This offset parameter hasn't been used in years and never worked reliably, either. Best to just drop it now, for simplicity; it's an unnecessary added complication. I'm going to simply delete out the code for this in 1., since that's going to be a not-backwards-compatible release anyway.

PR #385, opened 2021-01-05
Remove unused offset parameter to Detector class

mperrin:
Once upon a time there was an idea to allow off-centered Detector instances, so you could center a detector on a PSF that's tilted off to the side. The code for this never got fully working right (See #40), and there's been a failing unit test for it marked `xfail` for years in fact. Time has made clear that idea was half baked at best, that isn't functionality anybody needs - and so it's best to just delete out that parameter entirely and the associated code. Cleaner to get rid of unnecessary and non-working code rather than keep it around in the vague hope of maybe eventually needing and fixing it.


shanosborne (approved):
Seems straightforward enough and all the tests are passing, so looks good to me!

PR #386, opened 2021-01-05
Implement Scalar OPD optic; and update syntax for get_optical_system removing underscore.

mperrin:
This implements a very simple (but currently missing from poppy) optical element class for a simple scalar OPD value.

In most cases this isn't interesting to add (since just changing the total path length or piston has no effect on a PSF) but it can be useful as a building block for making compound optical elements. This PR also updates the CompoundAnalyticOptic class to correctly handle OPDS for compound optics with multiple apertures.  For instance here's a classical two slit interferometer. Adding half a wave of delay to one side makes for destructive interference on-axis, as expected: 

```python

wavelen = 500*u.nm
plt.figure(figsize=(20,8))
for i, delay_waves in enumerate([0, 0.5]):
    slit1 = poppy.RectangleAperture(width=50*u.micron, height=600*u.micron, shift_x = 100e-6)
    slit2 = poppy.RectangleAperture(width=50*u.micron, height=600*u.micron, shift_x = -100e-6)
    slit2_with_delay = poppy.CompoundAnalyticOptic((slit2,
                                                  poppy.optics.ScalarOpticalPathDifference(opd=delay_waves*wavelen)))
    slits = poppy.CompoundAnalyticOptic([slit1, slit2_with_delay], name='Two slits', mergemode='or')



    osys = poppy.FresnelOpticalSystem(pupil_diameter=1000*u.micron, npix=512, beam_ratio=0.25)
    osys.add_optic(slits)
    osys.add_detector(pixelscale=3*u.micron/u.pixel, fov_pixels=1024, distance=10*u.cm)
    psf = osys.calc_psf(wavelength=500*u.nm)
    
    plt.subplot(1,2,1+i)
    poppy.display_psf(psf, title=f'Two slits, with OPD={delay_waves} waves')
    plt.axvline(0, ls=':', color='white')
plt.figure()
slits.display(what='both')

```

Unknown

Unknown-1

This PR adds the optic itself. I'll include it in the documentation with this example code in a subsequent PR. 

**Note added later:** This PR also updates the syntax for the Instrument.get_optical_system function, which originally was a private function with leading underscore but for a while now has been intended as a public part of the API. This was intended to be a transparent update just flipping the sign of the back compatibility wrapper, but it turns out to be a little more complex and needs some updates in downstream user packages including webbpsf and hicat package.

### DIFF
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -53,7 +53,7 @@ class Instrument(object):
 
     You will at a minimum want to override the following class methods:
 
-        * _get_optical_system
+        * get_optical_system
         * _get_filter_list
         * _get_default_nlambda
         * _get_default_fov
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -259,7 +259,7 @@ class Instrument(object):
 
         # ---- now at last, actually do the PSF calc:
         #  instantiate an optical system using the current parameters
-        self.optsys = self._get_optical_system(fov_arcsec=fov_arcsec, fov_pixels=fov_pixels,
+        self.optsys = self.get_optical_system(fov_arcsec=fov_arcsec, fov_pixels=fov_pixels,
                                                fft_oversample=fft_oversample, detector_oversample=detector_oversample,
                                                options=local_options)
         self._check_for_aliasing(wavelens)
@@ -759,7 +759,7 @@ class Instrument(object):
         # (specifically auto-selected pupils based on filter selection)
         wavelengths, _ = self._get_weights(nlambda=1)
         self._validate_config(wavelengths=wavelengths)
-        optsys = self._get_optical_system()
+        optsys = self.get_optical_system()
         optsys.display(what='both')
         if old_no_sam is not None:
             self.options['no_sam'] = old_no_sam
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -498,7 +498,7 @@ class Instrument(object):
         """
         pass
 
-    def _get_optical_system(self, fft_oversample=2, detector_oversample=None, fov_arcsec=2, fov_pixels=None,
+    def get_optical_system(self, fft_oversample=2, detector_oversample=None, fov_arcsec=2, fov_pixels=None,
                             options=None):
         """ Return an OpticalSystem instance corresponding to the instrument as currently configured.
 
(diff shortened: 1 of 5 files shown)


poppy/instrument.py
```diff
@@ -259,7 +259,7 @@ def calc_psf(self, outfile=None, source=None, nlambda=None, monochromatic=None,
 
         # ---- now at last, actually do the PSF calc:
         #  instantiate an optical system using the current parameters
-        self.optsys = self._get_optical_system(fov_arcsec=fov_arcsec, fov_pixels=fov_pixels,
+        self.optsys = self.get_optical_system(fov_arcsec=fov_arcsec, fov_pixels=fov_pixels,
```
mperrin, line 262: Note, this isn't directly related to this PR's main idea, but is an overdue update to some syntax that might as well get included here.

poppy/optics.py
```diff
@@ -347,6 +347,20 @@ def get_transmission(self, wave):
...
+    """Uniform and constant optical path difference
+
+    """
+    @utils.quantity_input(opd=u.meter)
+    def __init__(self, name=None, opd=1.0*u.micron, **kwargs):
+        AnalyticOpticalElement.__init__(self, name=name, **kwargs)
+        self.opd = opd
+
+    def get_opd(self, wave):
+        res = np.empty(wave.shape, dtype=_float())
```
shanosborne, line 360: Did you mean `dtype=float` here? The same `_float()` call is also present on line 1892, but it doesn't seem to work when I try it locally with `np.empty(a.shape, dtype=_float())` where `a` is an array.

shanosborne (commented):
You could also update `_get_optical_system` in the docstring of the main `Instrument` class and the calls to `_get_optical_system()` in `test_instrument.py` for cleanliness, but you don’t have to since it’s just an alias. It also might be good to add a test for `ScalarOpticalPathDifference`, but again it’s not absolutely necessary.

poppy/optics.py
mperrin: no, the `_float()` with leading underscore is intentional. That's a little utility function defined in `accel_math.py`:
```
def _float():
    """ Returns numpy data type for desired precision based on configuration """
    # How many bits per float to use?
    return np.float64 if conf.double_precision else np.float32
```

shanosborne: Ahh it's from accel_math.py, okay that makes sense!

shanosborne (approved):

PR #388, opened 2021-02-04
MNT: Port from pysynphot to stsynphot

### LINKED ISSUE #362: Migrate from pysynphot to synphot
Synphot is a refactor of pysynphot and we should start using the newer package. This is recent work done by @ariedel in support of the ETC (see discussion at [this innerspace page](https://innerspace.stsci.edu/pages/viewpage.action?spaceKey=JWST&title=2020-06-15+ETC+release+schedule+update%2C+new+template+spectra+and+filter+bandpasses%2C+synphot+refactor) and [this presentation by Adric](https://innerspace.stsci.edu/download/attachments/220693357/Synphot_Presentation.pptx?version=1&modificationDate=1592254666182&api=v2) and [this Pandeia PR](https://github.com/spacetelescope/pandeia/pull/4949).

We should update poppy (and webbpsf) to also use synphot instead of pysynphot.

pllim:
Fix #362 

cc @mperrin 

**TODO**

- [ ] Have stakeholders review and test this.
- [x] Fix code and tests, if necessary.


poppy/instrument.py
```diff
@@ -852,27 +855,30 @@ def _get_weights(self, source=None, nlambda=None, monochromatic=None, verbose=Fa
...
+        Uses stsynphot (if installed), otherwise assumes simple-minded flat spectrum
 
         """
         if nlambda is None or nlambda == 0:
             nlambda = self._get_default_nlambda(self.filter)
 
         if monochromatic is not None:
             poppy_core._log.info("Monochromatic calculation requested.")
-            return (np.asarray([monochromatic]), np.asarray([1]))
+            return ([monochromatic] * units.AA, np.asarray([1]))
```
pllim: Ops, I tried something but forgot to revert.

```suggestion
            return (np.asarray([monochromatic]), np.asarray([1]))
```

poppy/instrument.py
```diff
@@ -886,38 +892,43 @@ def _get_weights(self, source=None, nlambda=None, monochromatic=None, verbose=Fa
...
-            poppy_core._log.debug("Min, max wavelengths = %f, %f" % (minwave / 1e4, maxwave / 1e4))
+            minwave = band_wave[w_above10].min()
+            maxwave = band_wave[w_above10].max()
+            poppy_core._log.debug("Min, max wavelengths = %f, %f" % (
+                minwave.to_value(units.micron), maxwave.to_value(units.micron)))
 
             wave_bin_edges = np.linspace(minwave, maxwave, nlambda + 1)
             wavesteps = (wave_bin_edges[:-1] + wave_bin_edges[1:]) / 2
             deltawave = wave_bin_edges[1] - wave_bin_edges[0]
+            jwst_area = 25.4 * (units.m * units.m)
```
pllim: `stsynphot` no longer let you set-and-forget the area like PySynphot does, because I deemed the risk of forgetting and then silently using the wrong value to be not worth the convenience.

poppy/utils.py
```diff
@@ -1601,32 +1604,32 @@ def sort_sptype(typestr):
 
     if "Flat" in sptype:
         if sptype == "Flat spectrum in F_nu":
-            spec = pysynphot.FlatSpectrum(1, fluxunits='fnu')
+            spec = SourceSpectrum(ConstFlux1D, amplitude=1 * syn_u.FNU)
         elif sptype == "Flat spectrum in F_lambda":
-            spec = pysynphot.FlatSpectrum(1, fluxunits='flam')
-        spec.convert('flam')
-        return spec * (1. / spec.flux.mean())
```
pllim, line 1608: I don't know why it is being normalized here. In the refactored package, flat spectrum is analytic and wavelength to evaluate it over isn't known ahead of time, so it is quite hard to normalize like this.

poppy/utils.py
```diff
@@ -1601,32 +1604,32 @@ def sort_sptype(typestr):
...
+        return spec
     if 'Power law' in sptype:
         import re
-        ans = re.search('\((.*)\)', sptype)
+        ans = re.search(r'\((.*)\)', sptype)
         if ans is None:
             raise ValueError("Invalid power law specification cannot be parsed to get exponent")
         exponent = float(ans.groups(0)[0])
-        # note that Pysynphot's PowerLaw class implements a power law in terms of lambda, not nu.
+        # note that synphot's PowerLaw class implements a power law in terms of lambda, not nu.
```
pllim, line 1618: It is now possible to define powerlaw in frequency space but that functionality isn't widely tested.

pllim (commented):
Someone needs to test this manually as I don't think your CI currently run the tests that requires PySynphot.

pllim:
Update: I got the tests to pass locally, but I don't know how good is your test coverage, so please review this PR carefully.

p.s. Would be nice if your package uses Quantity throughout; it defeats the purpose of using Quantity to only manually convert the units back and forth; but that is out of scope of this PR.

pllim:
Unrelated to this PR but I see this warning on your test page: `Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information.`

So, you either need to move to their `.com` service and start paying, or port your testing infrastructure to GitHub Actions. I see that you already use `tox`, so doing the latter should be very easy for you.

mperrin:
Thanks @pllim! @shanosborne can you give this a look, please. 

@pllim: on Travis yes we're aware, and @shanosborne has been coordinating on that with others (ideally this should be solved consistently for many/most of the INS packages, not on a case-by-case basis)

About quantities, agreed in principle. However, the quantity framework becomes somewhat complicated to use with Fourier transforms (which is the core of what this package does...), and further there were noticeable performance impacts if I tried to use quantity throughout in one test PR. Admittedly that was a while ago (a couple years).  It's possible some of this could be improved, but I agree it's out of scope for this PR

pllim:
Re: Performance of Quantity -- Might worth checking out https://docs.astropy.org/en/latest/units/index.html#performance-tips and see if it is of any use to you.

shanosborne (commented):
This PR looks good, thanks @pllim! We can't merge it right now because we've determined that the current dev version of `WebbPSF` can't run with PR and needs some updates (basically the same `pysynphot`->`stsynphot` updates). So we'll merge this PR once we have a comparable PR in `WebbPSF` ready to go.

pllim:
@shanosborne , not sure what you mean, as I developed this PR against your default branch and it is not showing that there is a conflict, but however you want to do this is fine by me. I do recommend that you test the changes extensively downstream before accepting the changes. Thanks!

shanosborne:
@pllim You were able to run the development version of `webbpsf` with the version of `poppy` in this PR? I agree that this version of poppy runs fine on it's own, but the downstream `webbpsf` package fails for me with this `poppy` change with the error of `AttributeError: 'ArraySpectralElement' object has no attribute 'waveset'`.

pllim:
Ah, sorry, I missed the fact that WebbPSF is a separate repository!

Hard to tell without the full traceback. What would be useful is if you tell me where the corresponding WebbPSF PR is and I can look at its CI log. Thank you!

conftest.py
```diff
@@ -0,0 +1,26 @@
+# this contains imports plugins that configure py.test for astropy tests.
```
mperrin, line 1: What's the story with this added contest.py file? This is pretty cryptic in parts, and I'd like to understand whether this is needed for some of the tests of synphot, whether it provides other benefits, and/or whether this causes any other changes in the behavior of the test system. 

I'd like to request another sentence or two of comments here to clarify what this stuff is for, and/or simplifying if any of this is not strictly needed. 

And also, is the package root directory really the place this file should be? I think in that case it does not get installed into site-packages since it's not within the `poppy` package subdirectory, right? Is that the intended behavior?

poppy/instrument.py
```diff
@@ -886,38 +893,54 @@ def _get_weights(self, source=None, nlambda=None, monochromatic=None, verbose=Fa
...
```
mperrin: We can simply delete this line with the jwst_area and remove the usage of it below. poppy is not interested in absolute fluxes, just normalized PSFs, so we can simplify and avoid hard-coding in this parameter with no change in results.

poppy/instrument.py
```diff
@@ -927,13 +950,13 @@ def _get_weights(self, source=None, nlambda=None, monochromatic=None, verbose=Fa
...
+        else:  # Fallback simple code for if we don't have stsynphot.
             poppy_core._log.warning(
-                "Pysynphot unavailable (or invalid source supplied)!   Assuming flat # of counts versus wavelength.")
+                "stsynphot unavailable (or invalid source supplied)! Assuming flat # of counts versus wavelength.")
             # compute a source spectrum weighted by the desired filter curves.
-            # The existing FITS files all have wavelength in ANGSTROMS since that is the pysynphot convention...
+            # The existing FITS files all have wavelength in ANGSTROMS since that is the stsynphot convention...
             filterfile = self._filters[self.filter].filename
-            filterfits = fits.open(filterfile)
+            filterfits = fits.open(filterfile)  # TODO: Use context manager
```
mperrin: This #TODO could be done now simply by replacing this line and the next with `filterdata = fits.getdata(filterfile)`, I think.

mperrin (approved):
This looks good. @shanosborne I have a few (minor, optional) suggestions below. 

One general question: has anyone done yet any performance benchmarking? I'm not expecting major speed changes either way, I just want to verify at a basic level there aren't any surprises. 

Side topic, I see this includes some better usage of temp directories in some of the unit tests - thanks, that's needed to be done for a while.

mperrin:
Oh and I see we now have some conflicts that need to be resolved before this is mergeable.

conftest.py
pllim: Story: The way `tox` is being run, `pytest` is not seeing (some of?) your `poppy/conftest.py`, so the test header isn't being displayed correctly in the CI logs (go see for yourself). This fixes the header.

Also see https://github.com/astropy/pytest-astropy-header/issues/15#issuecomment-763247031 .

As for how you want to document this story in a way that you understand, feel free to use the "suggestion" feature. I have been doing this for so long that I might not document it the way that make sense to you.

pllim (commented):
It is my understanding that @shanosborne has taken over the PR, so I'll let her handle the code pushes to address the reviews. I will be available for consultation. Please let me know if you have further concerns or questions. Thanks!

poppy/instrument.py
shanosborne: So `area` is a required parameter for `synphot.Observation`, but we can change it so it just takes in 1 m^2 rather than a specific number related to JWST.

pllim: `area` is required but it is up to you what value to give it.

In PySynphot, it was also required but PySynphot just automatically used HST area unless you overwrite the setting.

shanosborne:
@mperrin I've done the updates you requested and fixed the merge conflicts. I'll do some bench-marking next, but will likely not get it done today, just a heads up!

shanosborne:
@mperrin Quick bench marking results. I ran the following default calculation several times, both with the code in this PR and the code in the current master branch (confirming that `stsynphot` and `pysynphot` were called respectively):
```
import poppy
inst = poppy.Instrument()
%timeit inst.calc_psf()
```

My average results:
STSynphot Case: `3.08 s ± 103 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
Pysynphot Case: `2.86 s ± 51 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

mperrin:
@shanosborne Thanks. Looks like stsynphot is perhaps a little slower than pysynphot  (for whatever reason), but in any case not drastically different. Those two measurements are consistent at ~1 sigma in any case. @pllim that sound right to you? 

I just wanted to check this to verify no surprises. We want to move to stsynphot in any case so I will go ahead and merge this

pllim:
> Looks like stsynphot is perhaps a little slower than pysynphot

I agree with @mperrin that they are consistent within uncertainty. But if you really want to drill down on it, you can do a more detailed profiling and see where the bulk of the processing time is in both. `stsynphot` does include overhead of `astropy` units and models, but it also does not sample until needed, so it really depends on how you are using it.

PR #389, opened 2021-02-05
v1.0 pre-release branch

mperrin:
Release notes etc can go here. 

I may also opt to merge the sign conventions stuff into this branch prior to merging to develop, for reasons having to do with setup tools scm versioning (and not breaking downstream code currently using poppy develop, in particular the optics lab HiCAT testbed, until that can be more carefully coordinated)


mperrin:
These notes are getting moved into 0.9.2, for reasons described in #392

PR #392, opened 2021-02-09
Release 0.9.2

mperrin:
As discussed in slack, let's have a v0.9.2 release: 
> for poppy + webbpsf releases here is my thinking currently. It’s turned into a bit of a tangle given the interlinked dependencies between these two (and the optics lab code on the research side).
I’m going to make an v0.9.2 release branch of poppy containing the current code - but backing out the change in _get_optical_system syntax that I had meant to be backwards compatible but turned out not to be. I will make sure that both the released webbpsf 0.9.1 and current develop run correctly on that brach of poppy. That can be a poppy release v0.9.2 but does not need a webbpsf release, I don’t think.
That will give a stable point right now, so the optics lab code can continue with that for the time being, and then on top of that I can then finish up the 1.0 release and sign conventions change without disrupting anything in the lab
(which eventually wants that update, but I have enough moving pieces right now I want to treat it as a separable problem rather than having to deal with everything all at once)

Toward that end I:
* Copied over and updated the release notes that had been originally drafted for v1.0
* Reverted the back-incompatible change in the get_optical_system syntax
* Tested that both the released webbpsf v0.9.1 and current develop webbpsf pass their own tests on this branch. That's true and they do.

### DIFF
--- a/docs/relnotes.rst
+++ b/docs/relnotes.rst
@@ -5,14 +5,14 @@ Release Notes
 
 For a list of contributors, see :ref:`about`.
 
-1.0.0
+0.9.2
 -----
 
-.. _rel1.0.0:
+.. _rel0.9.2:
 
-*2021 Jan TBD*
+*2021 Feb 10*
 
-Release summary sentence here. 
+This release includes several updated optical element classes, bug fixes, and improved documentation. This is intended as a maintenance release shortly before v 1.0 which will introduce some backwards-incompatible changes. 
 
 **New Functionality:**
  * New OpticalElement classes for ScalarOpticalPathDifference, LetterFAperture, and LetterFOpticalPathDifference. (:pr:`386` by :user:`mperrin`)
@@ -36,8 +36,6 @@ Release summary sentence here.
  * Minor doc string clarification for get_opd (:pr:`381` by :user:`douglase`)
  * Remove unused parameter to Detector class (:pr:`385` by :user:`mperrin`)
 
-
-ADD fanpeng-kong, spacegal-spiff TO CONTRIBUTORS
 
 
 0.9.1
--- a/docs/relnotes.rst
+++ b/docs/relnotes.rst
@@ -10,7 +10,7 @@ For a list of contributors, see :ref:`about`.
 
 .. _rel0.9.2:
 
-*2021 Feb 10*
+*2021 Feb 11*
 
 This release includes several updated optical element classes, bug fixes, and improved documentation. This is intended as a maintenance release shortly before v 1.0 which will introduce some backwards-incompatible changes. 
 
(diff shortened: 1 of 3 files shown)


docs/relnotes.rst
```diff
@@ -5,6 +5,39 @@ Release Notes
...
+ * The ShackHartmannWavefrontSensor class was refactored and improved . (:pr:`369` by :user:`fanpeng-kong`). And a unit test case for this class was added (:pr:`376` by :user:`remorgan123` in collaboration with :user:`douglase`)
+ * Expanded documentation and example code for usage of astropy Units. (:pr:`374`, :pr:`378` by :user:`mperrin`; with thanks to :user:`keflavich’ and  :user:`mcbeth`)
+ * Bug fix for influence functions of rotated hexagonally segmented deformable mirrors. (:pr:`371` by :user:`mperrin`)
+ * Bug fix for FWHM measurement on integer data type images. (:pr:`368` by :user:`kjbrooks`)
+ * Bug fix for StatisticalPSDWFE to avoid side effects from changing global numpy random generator state. (:pr:`377` by :user:`ivalaginja`)
+ * Bug fix for image display in cases using angular coordinates in units other than arc seconds. (:pr:`378` by :user:`mperrin`; with thanks to :user:`mcbeth`)
+
+
+**Software Infrastructure Updates and Internals:**
+ * The minimum numpy version is now 1.16. (:pr:`356` by :user:`mperrin`)
```
shanosborne, line 33: PR 356 was for changing the minimum python version to 3.6 (i think the numpy comment if left over from a copy/paste)

shanosborne (commented):
I've confirmed that this version of poppy works fine with the stable branch of webbpsf, but I'm seeing the tests on the develop branch of webbpsf are failing with this branch.

docs/relnotes.rst
mperrin: Aha, looks like this was actually your PR https://github.com/spacetelescope/poppy/pull/367 which updated the numpy version. But in fact it looks like that only updated the CI test settings in `tox.ini` to have 1.16, and we still have 1.13 in the `setup.py` file. Seems like we should make that consistent.

In fact it seems like we have three sets of version information right now? Getting complicated. THere's 
 - requirements.txt, for creating environments with up-to-date packages, and use by dependabot for testing new versions
 - setup.py for setting minimum versions and allowing flexibility to people installing this in different environments
 - tox.ini for setting which versions get used on the CI system, including multiple options. 

I don't see any easy way to simplify that but it's not exactly simple. Can you do a quick check through those for any other inconsistencies, please?

shanosborne: Aha yes you're right. I've just gone through the files and updated an inconsistencies with a new commit in #395

mperrin:
>  I'm seeing the tests on the develop branch of webbpsf are failing with this branch.

Can you post a copy of the test log with the failures here? For me this is working correctly.  It's possible you may be seeing failures for an unrelated reason having to do with expected changes in data files.

shanosborne:
[webbpsf-test-output.txt](https://github.com/spacetelescope/poppy/files/5959847/webbpsf-test-output.txt)
Here's my terminal output, I confirmed I'm on the develop branch of webbpsf (and it's up to date) and on this branch of poppy, and I'm getting `AttributeError: 'NIRCam' object has no attribute '_extra_keywords'` on the tests

mperrin:
I'm baffled by that terminal output; those look like the right git commit IDs, and match what I have here locally myself, but for me this is all passing. 

I reverted the change in the get_optical_system syntax in this branch; so when run with webbpsf develop this should be using a call stack like this :
```
FGS.calc_psf = webbpsf.JWInstrument.calc_psf()
-> calls webbpsf.SpaceTelescopeInstrument.calc_psf(), which is the inherited poppy.Instrument.calc_psf()
     -> calls poppy.Instrument._get_optical_system()
     -> calls webbpsf.SpaceTelescopeInstrument.get_optical_system()
          # and that defines the _extra_keywords among its many other steps
```
I have verified the above is what happens when I run this. 


The problem occurs when instead the following happens on poppy v0.9.1 and webbpsf develop: 
```
FGS.calc_psf = webbpsf.JWInstrument.calc_psf()
-> calls webbpsf.SpaceTelescopeInstrument.calc_psf(), which is inherited poppy.Instrument.calc_psf()
     -> calls poppy.Instrument._get_optical_system(), which just runs directly and doesn't call the subclassed version.
```

The only way I can recreate the behavior you are seeing is if I install poppy 0.9.1 and run it with webbpsf develop...

mperrin:
@Skyhawk172 has also tested this branch with webbpsf develop and has the same result as me (it all works). Meanwhile @shanosborne has retested and still has the same result she did before (it doesn't work). This is curious, to say the least. @shanosborne will continue investigation tomorrow.

shanosborne (approved):
Upon further investigation, the problem was conda related and not poppy related. So this PR is good to go!

PR #394, opened 2021-02-09
runtime improvement by avoiding rotating zero-padding

mperrin:
When rotating wavefronts, if they're zero-padded don't bother rotating all those zeros. 

Addresses downstream issue in webbpsf: https://github.com/spacetelescope/webbpsf/issues/406. Provides a substantial improvement in JWST MIRI simulations (reduces runtime by 50%).

Relatedly, improve the debugging option `conf.enable_speed_tests` to debug log the calculation runtime for each optical element in an optical system, not just the overall runtime per wavelength. This was useful in pinning down that it really was the rotation step that was taking so much time.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -1142,7 +1142,7 @@ class FresnelOpticalSystem(BaseOpticalSystem):
 
         for optic, distance in zip(self.planes, self.distances):
 
-            if poppy.conf.enable_speed_tests:
+            if poppy.conf.enable_speed_tests:  # pragma: no cover
                 s0 = time.time()
 
             # The actual propagation:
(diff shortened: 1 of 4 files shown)


douglase:
It might also be possible to speed things up by switching to nearest neighbor interpolation. We (h/t @jlumbres) also been having some issues with artifacts from higher-order spline rotation. 

SciPy seems to be pushing folks towards using transforms in skimage (https://docs.scipy.org/doc/scipy-1.1.0/reference/generated/scipy.misc.imrotate.html#scipy.misc.imrotate) and in a quick test skimage rotation is ~2x faster on my machine, which is interesting since it looks like they are both just using one CPU core: 
Screen Shot 2021-02-09 at 3 23 28 PM

mperrin:
@douglase and @jlumbres - I'm curious if you have done any benchmarking of the numerical precision of `scipy.ndimage.rotate` vs `skimage.transform.rotate`. I can replicate your benchmark results that the skimage version is 2-3x faster. However it also appears noticeably less accurate too: 

```python
import poppy
from scipy.ndimage import rotate
import skimage.transform
import copy


# Set up two optical systems, one with an aperture rotation and one without
osys = poppy.OpticalSystem()
osys.add_pupil(poppy.RectangleAperture(rotation=0))
osys.add_detector(0.05, fov_pixels=128)

ang = 5.5
osysr = poppy.OpticalSystem()
osysr.add_pupil(poppy.RectangleAperture(rotation=ang))
osysr.add_detector(0.05, fov_pixels=128)

# Calculate PSFs
psf = osys.calc_psf()
psfr = osysr.calc_psf()

fig, axes = plt.subplots(ncols=2, figsize=(10,3))
poppy.display_psf(psfr, ax=axes[0], title="Reference PSF\nfrom rotated rectangle aperture")
poppy.display_psf(psf, ax=axes[1], title="PSF to transform\nfrom non-rotated rectangle aperture")

# Compare the two rotation methods
for method in ['scipy', 'skim age']:

    fig, axes = plt.subplots(ncols=2, figsize=(10,3))

    p0r = copy.deepcopy(psf)
    if method=='scipy':
        p0r[0].data = rotate(psf[0].data, angle=-ang, reshape=False, order=3)
    else:
        p0r[0].data = skimage.transform.rotate(psf[0].data, angle=-ang, order=3)
    poppy.display_psf(p0r, title=f'PSF rotated with {method}', ax=axes[0])

    x0,x1 = 50,210
    y0,y1 = 50,210

    diff = psfr[0].data[y0:y1, x0:x1] - p0r[0].data[y0:y1, x0:x1]

    norm_rms = diff.std() / psfr[0].data.max()
    poppy.display_psf_difference(psfr, p0r, vmax=1e-6, 
                                 title=f'Difference using {method}: $\sigma$={norm_rms:.3e}\n',
                                ax=axes[1])
```


Screen Shot 2021-02-22 at 12 06 47 PM

mperrin:
On further investigation, results from the two algorithms are identical for order=0 or 1, then diverge for higher orders, with the scipy.ndimage version always providing lower error in this particular test case.

shanosborne (approved):
This looks good to me! If we wanted, we could take this opportunity to replace `utils.removePadding` with `utils.remove_padding` to make it more Pythonic. But that may also fall into a backward compatibility hole, so there's no reason we have too.

mperrin:
Thanks! Let's make it a separate PR for the PEP8ing of remove padding. But good idea on that, yes.

PR #398, opened 2021-02-23
Use Intel MKL FFT, if available

mperrin:
@ehpor informed me that the Intel MKL FFT is no longer automatically patched in as a replacement for numpy.fft; you have to call mkl_fft specifically. This does so, for significant performance gains. 

We should add some mention to the docs performance section to recommend installing Intel MKL FFT; and more generally the performance discussions are in several placed outdated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
         - env: TOXENV='py38-cov'
 
         # Check accelerated math version too
-        - env: TOXENV='py37-numexpr-cov'
+        - env: TOXENV='py37-numexpr-mkl-cov'
 
         # Check for Sphinx doc build errors
         - env: TOXENV='docbuild' TOX_ARGS=''
--- a/docs/fft_optimization.rst
+++ b/docs/fft_optimization.rst
@@ -1,6 +1,11 @@
 Appendix B: Optimizing FFT Performance with FFTW
 ================================================
 
+.. warning::
+
+    This page is obsolete, and superceded by subsequent library development. You probably want
+    to use Intel MKL rather than FFTW. See :ref:`_performance_and_parallelization`.
+
 Optimizing numerical performance of FFTs is a complicated subject. Just using the FFTW library is no guarantee of optimal performance; you need to know how to configure it.
 
 .. note::
--- a/docs/performance.rst
+++ b/docs/performance.rst
@@ -5,21 +5,33 @@ Appendix A: Optimizing Performance and Parallelization
 
 Performance optimization on modern multi-core machines is a complex subject.
 
+.. note::
+
+   TL, DR: You probably want to have `a version of numpy using Intel MKL, such as the one in Anaconda <https://docs.anaconda.com/mkl-optimizations/index.html>`_,
+   plus the Intel `mkl_fft <https://anaconda.org/intel/mkl_fft>`_ library, plus `numexpr <https://github.com/pydata/numexpr>`_.
+
+
 Know your Linear Algebra Library
 ------------------------------------
 
-The optical propagation calculations in POPPY are dominated by a small number
-of matrix algebra calls, primarily ``np.dot`` with a side order of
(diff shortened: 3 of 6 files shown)


mperrin:
@shanosborne two requests for this one. First and foremost can you please review this PR soon? It's brief and straightforward. 

Secondly I don't understand what's going on with the code coverage tests. I added MKL to one of the Travis test builds, so it should be being included and tested. But codecov doesn't like this PR despite that. I try to click through and get details but I end up getting cryptic errors from the codecov website. Try this and see if you can make it work: 
https://codecov.io/gh/spacetelescope/poppy/src/b5c095937f70e35f617a47c0cdaca21124491c2b/poppy/accel_math.py
I am not sure what's going on. I don't want to waste time fighting it... but it sure does seem something is not as it should be. 

While I'm thinking about CI issues... do we have any long term guidance yet for the Travis migration/shutdown/etc?

shanosborne:
So I with some investigating, I think codecov is actually right and those lines weren't getting tested by travis. I looked into `mkl_fft` and the version on pypi only works for python <=3.6, plus lots of people (including me) have problems installing it. So while travis says that it got installed, I'm not sure it actually did (e.g. when I run the same tox test locally, it doesn't install). So I messed around with the tox file and I think I've got a version of it that actually installs the package (now using conda).

Also, I can only access that link once I'm logged into codecov (in my case, I log in via my GitHub account). So I'd check that you're logged in

I've confirmed that when I run this locally, the relevant lines in accel_math.py are now getting hit. But for some reason it's not translating in to CodeCov. I'm looking into it
```
<line hits="1" number="248"/>
<line hits="1" number="250"/>
<line hits="1" number="251"/>
<line hits="1" number="252"/>
<line hits="1" number="253"/>
```

mperrin:
Thanks @shanosborne. In fact I hadn't realized or intended that I'd put mkl_fft in the pip rather than conda section. I don't think I had appreciated the difference between `deps` and `conda_deps` for tox.  I wonder if we should also move the numexpr dependence to conda as well?

I'm logged into codecov now, and can see some more of the files, but in other ways it's still fighting me. For example if I go to https://app.codecov.io/gh/spacetelescope/poppy/compare/398/changes and try to view any of the changes, it just spins forever and doesn't show anything... weird.

mperrin:
Some FFT performance testing. MKL wins out over pyfftw, as expected, though it gets close at larger array sizes. Both easily beat the numpy implementation. 

image

Note, the FFTW timings are using (as poppy does) the FFTW_MEASURE planner level in FFTW, which squeezes out max performance based on empirical timings, and explicitly requesting multi-threaded FFTs. This is not the default behavior from FFTW out of the box. I mention this just because it can complicate benchmarking; a naive benchmark of FFTW can look much slower if it's not set to multi-threaded. Here's a similar chart showing comparisons of some of the lower-level FFT routines, including with/without threading for FFTW: 

image

shanosborne:
If numexpr needs to be installed with conda, then yes we can move it. But since tox is based in pip, and the conda installation method is an add on from an separate package, I feel like keeping as much in pip as possible is probably a good method for avoiding breaking things. 

And yeah that codecov page eternally spins for me too. But I've found that if you go over to the commits tab and choose a specific commit (like the most recent one), then you can click the files are they load. I have no idea why that method works, but it's better than nothing.

mperrin:
got it - I had not realized tox was more pip-based than conda-based. Makes sense given that explanation.

mperrin:
@shanosborne Are you planning to do any more work on the codecov or tox setup for this branch? I am inclined to just merge this as is, since the MKL speedup is significant, and deal with sorting out CI/coverage issues later.  This is especially the case given how the codecov website seems to be misbehaving for some reason.

shanosborne (approved):
Yeah I agree. I've just double checked that the tests are hitting the new code on my local runs, so this is a separate codecov issue. I think we're good to merge this PR!

PR #400, opened 2021-02-25
Change for loop calculating OPD from Zernikes to an Einstein sum.

grbrady:
I change a for loop in the opd_from_zernikes function in Zernikes.py to an Einstein sum.  This produced a significant speedup in code that I wrote (JWST field dependence model in WebbPSF) that uses the function heavily.  This employs the Numpy einsum function, which is a highly optimized, multithreaded implementation.  Possibly downside is that the Einstein sum notation is a little less readable to most people.

### DIFF
--- a/poppy/zernike.py
+++ b/poppy/zernike.py
@@ -1153,10 +1153,8 @@ def opd_from_zernikes(coeffs, basis=zernike_basis_faster, aperture=None, outside
     constant_support =  np.allclose(np.isfinite(basis_set[0]), np.isfinite(basis_set[-1]))
 
     if constant_support:
-        # we can just sum the whole arrays
-        for i, b in enumerate(basis_set):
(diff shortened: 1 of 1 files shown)


grbrady:
I can't seem to explicitly add a reviewer to this pull request.  Will follow-up with Marshall or Shannon as to whether this is a problem.

mperrin (approved):
👍 A nice optimization, thanks @grbrady 

Some benchmarks from calls to `opd_from_zernikes` show 2x or better speedup:  

| # zernikes | Npix |  using for loop | using einsum |
| ---------| -------| ---------------| --------------| 
| 36 | 512 |  21.7 ms | 10.8 ms | 
| 36 | 1024 |  116 ms | 49.1 ms | 
| 78 | 1024 |  235 ms | 87.4 ms |

PR #384, opened 2020-12-03
Sign convention clarification, update, disambiguation, and documentation.

mperrin:
Significant and wide-ranging changes to clarify and disambiguate sign conventions, following recent discussions with folks from the JWST WWG. 

See [this Jupyter notebook](https://github.com/mperrin/poppy-st/blob/sign_conventions/docs/sign_conventions_for_coordinates_and_phase.ipynb) for discussion, documentation, and test outputs to verify the sign conventions as updated. 

- The principle and most significant change is flipping the sign of the complex exponentials used in optical propagations, including both the FFT and MFT implementations. 
- This has effects that ripple out across poppy, necessitating associated sign changes in code for wavefront tilts, thin lenses, the FQPM_FFT_Aligner virtual optic, and existing tests thereof. 
- Unit test functions are added for several checks on sign conventions, in a new `test_sign_conventions.py`. 
- In support of some of these tests, I added a few new OpticalElement classes for basic scalar and tilt OPDs, a new parity test letter F aperture (which has been wanted for a while), and enhanced the Wavefront classes to have a .wfe attribute as well as .phase. (These are proportional, but WFE is in linear units equivalent to OPD and phase is in radians). 

- [x] Continue verification and validation of the above, in particular for the Fresnel propagation side of things. This is now updated and also verified consistent. 

**TODO:**
- [ ] Assess necessary changes on downstream packages including webbpsf, hicat, morphine, and others.

PR #402, opened 2021-02-27
Updates to fresnel.py allowing users to create their own inwave object.

kian1377:
The changes made are such that a user can input their own custom wavefront data into a FresnelOpticalSystem() when using calc_psf().

The methods calc_psf() and propagate_mono() have been added directly into the FresnelOpticalSystem() class, so they are not just inherited and they contain an additional kwarg that is inwave. This kwarg is initialized to be None, but the user can define their own FrenselWavefront() and input it into calc_psf() using the inwave kwarg. This inwave object is then passed to propagate_mono(), which subsequently passes inwave to the input_wavefront() method in FresnelOpticalSystem(), which has also been edited to initialize a wavefront if the value of inwave is None, or to raise a value error if something other than a FresnelWavefront() object was supplied.

The changes made have been in an effort to create the Roman CGI modes with POPPY, however, the SPC modes and their data files required a unique sequence of MFTs and FFTs to apply the Focal Plane Masks (FPMs). So two separate optical systems were created, the first going up to the FPM and the second going through the rest of the optics of the CGI. This way, the FPM could be applied to the final wavefront of the first optical system, and then input into the second system to continue propagation. The results of this system have been tested and show agreement with the results of the PROPER model.

results

Future efforts will include implementing the technique used to apply the FPMs directly into POPPY such that two separate optical systems do not need to be created, but having the flexibility for a user to input their own wavefront data still seemed useful.

Please let me know if there issues with the PR or anything else I should change. Others who have contributed are @douglase and @Jashcraf.


mperrin (changes requested):
Hi @kian1377, 

I think we should consider making these changes in the parent class `calc_psf` instead. As much as possible we would like to keep the same consistent API for both the Fresnel and Fraunhofer classes. Furthermore, right now this adds a lot of code duplication, which makes maintenance harder in the future. 

The inwave parameter is just as applicable to the Fraunhofer propagation class as well, yes?  So I would like to request revising this to instead make the changes in `poppy_core` to add that parameter, which will provide that option consistently for both propagation methods as well as avoiding the need to duplicate 300+ lines of code. 

Thanks!

kian1377:
Okay, I have been working on implementing it that way today and writing a couple tests to go along with those changes.

douglase (changes requested):
it looks like this commit creates some duplicate files that need to be moved back into their subdirectories

kian1377:
Yeah, I just noticed that after I updated my repo and committed the
changes. I'll fix that real quick.

mperrin:
Hmm, I'm confused what's going on with the changed files display in this PR now. What I see there appears to be a copy of some of the changes from PR #384, and I don't see anything about the inwave parameter. Not sure what's going on.

kian1377:
It looks like the tests that are failing are those for the compound systems, I will look into these later today and try and fix them.

kian1377:
I am taking a look at the changes made and I don't see anything about PR #384, I never even had anything to do with that so I don't know whats going on there either.

kian1377:
So all tests should be passing at this point because I ran the three tests related to the CompoundOpticalSystem's on my machine and they ran without error. Please let me know if there are more issues I need to resolve first.

mperrin:
Hi @kian1377, it does look like there is some work work to resolve on this PR. In particular if you look at the GitHub display it shows there are conflicts, like this:

**Conflicting files**
`poppy/poppy_core.py`

That will have to be resolved. The best way to do that is for you to  merge the current `develop` branch into your branch (i.e. `git merge upstream/develop`), resolve any conflicts, commit that change and then push to this branch. Thanks.

kian1377:
Hello, I will try and resolve the conflict within the next day. I am still getting used to github and don't fully understand pull requests and how to deal with different branches. I thought by updating the files in my repo, which I made the PR from, the PR would automatically update but it seems something went wrong this time.

kian1377:
Hello @mperrin , so I was trying the git merge upstream/develop command but I kept on running into issues with conflicts and had to jump through some hoops to update my local clone of poppy. In the end, I downloaded the zip code of spacetelescope/poppy develop branch and replaced all my files in my repo with those and committed the changes. Then, I made the edits to the poppy_core.py, frensel.py, and test_core.py files that implement the inwave kwarg and committed those changes as well. 

So now, there shouldn't be any conflicts and you should see the inwave updates.

mperrin:
Hi @kian1377, there are still a few changes and some cleanup I'd like to make in this. I have the changes locally here. It might be easiest if I can just push these directly to your branch, which I have tried to do but ran into a permissions issue. Can you please double-check that the GitHub setting "allow edits from maintainers" is set for this pull request, please? See https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

kian1377:
Hello, the "allow edits from maintainers" option is already selected. I am still looking into options that may be preventing your changes from being pushed in my forked repo of poppy.

kian1377:
I have unchecked an option for my forked repo saying "restrict editing to collaborators only." Maybe this was the option preventing the changes to be pushed directly to my branch. @mperrin when you get the chance, please try to push the changes again. 

If it still doesn't work, you could send me the files that you have changed via email at kianmilani@email.arizona.edu and I could push them to my repo.

kian1377:
@mperrin , I updated the PR with the files you sent me. I am not sure why git was giving us trouble with this but being new git and PRs, I'm pretty sure its something on my end that I just haven't figured out yet.

mperrin (approved):
Looks good. Thanks again Kian both for implementing this and for working through the revisions with me.

(And I'm not sure why git is being a pain on this one - it's still showing in the diff duplicate copies of a few changes already in develop, for some reason - but it's not a roadblock. Git is not the most user-friendly piece of software in the world, alas.)

PR #406, opened 2021-03-04
MNT: Fix and update license

pllim:
It is confusing to have 2 copies of similar license, so I picked the one that sounds more accurate and also update the year and include it in package manifest.


shanosborne (approved):
Looks good!

PR #407, opened 2021-03-16
Multi circular aperture

Teusia:
Dear all,

I needed to simulate a segmented mirror with circular segments. The modifications are heavily copy/pasted from the `HexSegmentedDeformableMirror` and the `MultiHexagonAperture` classes. I only made changes to account for the circular segments instead of hexagonal. 
I hit only one small snag: I used the `CircularAperture` class to draw the segments (again a copy paste) and you can grey scale pixels at the edge of the circular apertures. I tried to have this accessible from the dms but it yield an error in the geometry method. hence, for now the grey scaling is forced to not happen. I left a `FIXME?` in the code to mark where this is happening. 

I hope I did a good enough job and you find this addition interesting and consider it for the general poppy library.

Best Regards,
Anne-Laure Cheffot

### DIFF
--- a/poppy/dms.py
+++ b/poppy/dms.py
@@ -7,6 +7,7 @@ import scipy.ndimage.interpolation
 import scipy.signal
 import astropy.io.fits as fits
 import astropy.units as u
+from abc import ABC, abstractmethod
 
 from . import utils, accel_math, poppy_core, optics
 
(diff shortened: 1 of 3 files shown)


mperrin:
Thanks, this looks like a good contribution indeed. It may be a while before I can read through the code in detail but I will try to get you comments within the next week or so. Thanks!

mperrin:
Hi @Teusia, 

This is a great addition. There's a lot of duplicated code between the multi hex and multi circle classes. With your permission I'd like to refactor this to share code between those functions via subclassing. I have a working version of this here now: first I define a MultiSegmentAperture generic abstract base class, which has 90% of the code, then define subclasses MultiHexagonAperture and MultiCircularAperture which differ only in the geometry of the segments. 

With your permission I'd like to add that revision as commits pushed onto this branch - is that OK?

Teusia:
@mperrin Yes that is totally fine with me, please go ahead. 
I actually am glad you are doing the subclassing because I tried and failed horribly at it. 
I would then delete this branch?

mperrin:
@Teusia - I've updated this branch to add the version which uses subclassing. If you want to see how it works you can look a the diff on GitHub. 

The basic idea is I set up abstract classes for a segmented aperture and for a segmented deformable mirror, and then subclassed each of those to hexagons and circles. (It's a little subtle though because I needed to use multiple inheritance, so I'm not surprised you found it a bit tricky! But take a look if you want to learn how I did it)

Tests have passed and I'm going to go ahead and merge now. Thanks again for contributing!

mperrin (approved):

Teusia:
> (It's a little subtle though because I needed to use multiple inheritance, so I'm not surprised you found it a bit tricky! But take a look if you want to learn how I did it)

Definitively into the learning business, thank you!!!
And sorry it required so much work from your side.

PR #411, opened 2021-03-23
add rot90 optimization (fixes #408), and fix inconsistent sign bug in rotation

### LINKED ISSUE #408: add more efficient rotation for 90 degree cases
pointed out by @ehpor on https://github.com/spacetelescope/hicat-package/pull/556#issuecomment-802155424

> the rotations inside poppy are done with scipy.ndimage. That feels overkill for a 90degree rotation. Is there a better way in poppy to do a 90degree rotation (np.rot90())?

It would be easy to add a check for 90 degree rotation and use a more performant function. This should be added in.

mperrin:
- Adds a check for 90 degree rotation (for rotations of either a Wavefront or FITSOpticalElement) or multiple of 90 degrees, and if so uses the faster and more exact `np.rot90` rather than an interpolation
- Add new tests to test consistency of the exact 90 degree rotation and an 89.9999 deg rotation via the interpolation method. This demonstrates consistent signs of rotation (positive is counterclockwise following the usual astronomical convention)
- _Fixes a bug (sign inconsistency) that I hadn't realized existed_, in which positive rotations were clockwise for wavefronts. This behavior directly contradicted the doc string of Wavefront.rotation which states rotations are counterclockwise; see poppy_core.py line 869. (Not sure how this didn't get caught before; whoops!) I opted to fix the function behavior for consistency with the docs, as well as consistency with the other rotations in poppy. 
- Update test for rotations in an optical system, to reflect the correct behavior after the above bug fix. 
- Add test that rotating an optic as an analytic function vs as a FITS array yields consistent results.

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -3011,7 +3011,7 @@ class FITSOpticalElement(OpticalElement):
                     # this is imperfect at best, of course...
                     self.amplitude = scipy.ndimage.interpolation.rotate(self.amplitude, -rotation,  # negative = CCW
                                                                         reshape=False).clip(min=0, max=1.0)
-                    wnoise = np.where((self.amplitude < 1e-3) & (self.amplitude > 0))
+                    wnoise = (self.amplitude < 1e-3) & (self.amplitude > 0)
                     self.amplitude[wnoise] = 0
                     self.opd = scipy.ndimage.interpolation.rotate(self.opd, -rotation, reshape=False)  # negative = CCW
                 _log.info("  Rotated optic by %f degrees counter clockwise." % rotation)
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -880,8 +880,16 @@ class BaseWavefront(ABC):
         # Huh, the ndimage rotate function does not work for complex numbers. That's weird.
         # so let's treat the real and imaginary parts individually
         # FIXME TODO or would it be better to do this on the amplitude and phase?
-        rot_real = scipy.ndimage.interpolation.rotate(self.wavefront.real, angle, reshape=False)
-        rot_imag = scipy.ndimage.interpolation.rotate(self.wavefront.imag, angle, reshape=False)
+
+        k, remainder = np.divmod(angle, 90)
+        if remainder == 0:
+            # rotation is a multiple of 90
+            rot_real = np.rot90(self.wavefront.real, k=-k)  # negative = CCW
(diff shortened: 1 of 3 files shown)


mperrin:
FYI @ehpor

mperrin:
```
test_fits_rot90_vs_ndimagerotate_consistency(plot=True)
```
Unknown


```
test_wavefront_rot90_vs_ndimagerotate_consistency(plot=True)
```

Unknown-1

mperrin:
This will require downstream updates in webbpsf and hicat.

mperrin:
@shanosborne ping - did you see the review request on this, please

shanosborne:
So I was reading up on the np.rot90 function and it looks to me like using positive k values rotates the array CCW and negative k values is clockwise. I did a quick plot and this is what I got. You say in the code `rot_real = np.rot90(self.wavefront.real, k=-k) # negative = CCW`, but the input angle is also in degrees CCW. Am I missing a subtly here?
Screen Shot 2021-03-26 at 9 24 30 AM

mperrin:
The subtlety is where the image origin is!  Your plots have origin=upper left, but poppy always sets origin=lower left (as now stated explicitly in the sign convention and orientation docs, etc)

mperrin:
This is a good catch that I should clarify this in the comments there perhaps

shanosborne:
Ahh okay yes when I plot with origin=lower everything looks good again - thanks!

poppy/poppy_core.py
```diff
@@ -2991,13 +2999,21 @@ def __init__(self, name="unnamed optic", transmission=None, opd=None, opdunits=N
...
+                    # rotation is a multiple of 90
+                    self.amplitude = np.rot90(self.amplitude, k=-k)   # negative = CCW
+                    self.opd = np.rot90(self.opd, k=-k)
+                else:
+                    # arbitrary free rotation with interpolation
+                    # do rotation with interpolation, but try to clean up some of the artifacts afterwards.
+                    # this is imperfect at best, of course...
+                    self.amplitude = scipy.ndimage.interpolation.rotate(self.amplitude, -rotation,  # negative = CCW
+                                                                        reshape=False).clip(min=0, max=1.0)
+                    wnoise = np.where((self.amplitude < 1e-3) & (self.amplitude > 0))
```
shanosborne: You can also remove the np.where() call here if you like

shanosborne (approved):
Looks good!

Issue #412, opened 2021-03-24
faster to use bool array indexing than np.where when setting up optics


mperrin:
A small but significant performance optimization. Several of the optical element classes use `numpy.where`  for selecting array subregions, but it's much faster with numpy to use boolean array masks: 

```python
# setup
wf = poppy.Wavefront()
y, x = wf.coordinates()
r = poppy.accel_math._r(x,y)
array = np.zeros_like(r, float)

# time using where
%timeit array[np.where(r<3)]=1.0
7.1 ms ± 384 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# time using boolean masks directly
%timeit array[r<3]=1.0
834 µs ± 18.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

(The use of `where` like this is an unintended hold-over in early code from my prior work in IDL, in which `where` is a very frequent part of the idiom for performance optimization. But it's not optimal in Python!)

mperrin:
See #413

mperrin:
Done by #413

PR #413, opened 2021-03-24
faster array indexing via boolean masks. Addresses #412.

mperrin:
Low-level change in numpy array indexing syntax to use e.g. `array[blah is true]` rather than `w = np.where(blah is true); array[w]`. Speeds up a variety of tasks such as optic setup and sampling.  A small optimization for performance but adds up over time.

I didn't extensively benchmark every part individually, but some spot tests confirm noticeable speedups:

**current develop**:

```
%pylab inline
import poppy

ap = poppy.HexagonAperture()
%timeit ap.sample(npix=1024)
43.8 ms ± 1.51 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

ap = poppy.CircularAperture()
%timeit ap.sample(npix=1024)
45.5 ms ± 1.22 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

ap = poppy.SquareAperture()
%timeit ap.sample(npix=1024)
40 ms ± 1.28 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

**with this PR:**

```
ap = poppy.HexagonAperture()
%timeit ap.sample(npix=1024)
30.8 ms ± 768 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

ap = poppy.CircularAperture()
%timeit ap.sample(npix=1024)
40.6 ms ± 1.39 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

ap = poppy.SquareAperture()
%timeit ap.sample(npix=1024)
27.7 ms ± 1.3 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### DIFF
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -887,7 +887,7 @@ class Instrument(object):
             poppy_core._log.info("Computing wavelength weights using synthetic photometry for %s..." % self.filter)
             band = self._get_synphot_bandpass(self.filter)
             # choose reasonable min and max wavelengths
-            w_above10 = np.where(band.throughput > 0.10 * band.throughput.max())
+            w_above10 = (band.throughput > 0.10 * band.throughput.max())
 
             minwave = band.wave[w_above10].min()
             maxwave = band.wave[w_above10].max()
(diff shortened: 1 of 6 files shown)


shanosborne (approved):
Looks good!

PR #414, opened 2021-03-26
Tip tilt stage

mperrin:
Add optical element class for a tip tilt stage, that allows adjusting the tip and tilt of any arbitrary optic placed on the stage. I developed this a while ago for the HICAT lab testbed simulator, and it's worth moving here into poppy for use elsewhere. 

This PR adds that class, a unit test for the class, and adds example code and some figures to the available optics documentation page. While I was editing that docs page I also added docs for the segmented circular DM that was just added in PR #407. 

I think we can omit any formal review for this PR, since this code was already reviewed and approved once in the HICAT repo (https://github.com/spacetelescope/hicat-package/pull/325 though that's a private repo so not viewable to all).


mperrin:
The test failure is due to a units handling issue in an older version of astropy, and only fails in our py36-legacy-test version.  Rather than figuring out the right way to work around this, at this point I think it makes more sense just to update the min versions. See #415.

mperrin:
Tests pass. As noted above, no need for a review on this one since already reviewed elsewhere; code is routinely used and working in the lab.

PR #415, opened 2021-03-29
update minimum versions of dependencies

mperrin:
Update astropy, matplotlib etc. minimum required versions to the current major release version. 

For scipy and numpy I semi-arbitrarily adjusted the to versions from sometime in 2019, not from any particular need but just a sense that we may not need to support arbitrarily old versions indefinitely. poppy 1.0 is a good time to update these minimums.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -19,16 +19,17 @@ latest development version from git::
 Requirements
 --------------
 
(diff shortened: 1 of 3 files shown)


shanosborne (approved):
Looks good!

PR #416, opened 2021-03-31
Update installation docs

### LINKED ISSUE #410: poppy for windows users and git
Hello,

So I had to verify that my windows colleagues can use poppy, and it turns out that the setup.py of poppy requires git. The windows machine I tested on had git Desktop installed, which seam not recognised by python. Because I am using Anaconda to manage python I could solve this by an anaconda install of git  (e.g.: `conda install git` there is also a python-git in pypi). 
Maybe this would be worth mentioning in the installation instruction?
However is there really a fundamental need for git in the poppy installation?
I think this is somewhat important to mention because the most common stack overflow first answer for the error raised when poppy install fails (which is `OSError: 'git' was not found`), is to add a PYTHONPATH to windows and add the path to a installed version of git. This breaks Anaconda!!!

Cheers,
Anne-laure Cheffot

shanosborne:
This PR clarifies the documentation on how to install poppy. Closes #410

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -9,9 +9,7 @@ POPPY may be installed one of three different ways.
 
     % pip install poppy --upgrade
 
-3. Cloning the source code hosted in `this repository on GitHub
-<https://github.com/spacetelescope/poppy>`_. It is possible to directly install the
-latest development version using your locally installed ``git`` package::
+3. Cloning the source code hosted in `this repository on GitHub <https://github.com/spacetelescope/poppy>`_. It is possible to directly install the latest development version using your locally installed ``git`` package::
 
     % git clone https://github.com/spacetelescope/poppy.git
     % cd poppy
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -1,20 +1,20 @@
 Installation
 ==================
 
-POPPY may be installed from PyPI in the usual manner for Python packages::
+POPPY may be installed one of three different ways.
 
-   % pip install poppy --upgrade
+1. Using ``conda`` through the `AstroConda channel <https://astroconda.readthedocs.io/en/latest/>`__. Note that Astroconda does not support Windows.
```
mperrin: Let's say like: "This is the recommended channel for most users on MacOS and Linux. But note that AstroConda does not support Windows."

mperrin (approved):
Looks good thanks. One very minor suggestion; optional.

PR #420, opened 2021-04-01
improve algorithm for source offsets with coord rotations + inversions

mperrin:
PR #411 fixed an error or inconsistency in rotations. This turned up an issue with offsets for sources in optical systems with  rotations and inversions, which shows up in WebbPSF MIRI and NIRSpec sims. (I think this may have only worked before due to an even number of bugs cancelling out.)

I reworked the source offset implementation to work backwards through the transforms rather than forwards, which makes it simpler to handle multiple transforms in a row self-consistently. 

I also expanded the unit test for this to include cases for optical systems with no transforms, with an inversion, with a rotation, and both. 

All tests pass locally, for poppy and webbpsf.

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -2074,31 +2074,37 @@ class OpticalSystem(BaseOpticalSystem):
 
         if np.abs(self.source_offset_r) > 0:
             # Add a tilt to the input wavefront.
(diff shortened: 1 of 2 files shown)


shanosborne (approved):
As discussed on slack, I'm getting some failures on the webbpsf side when I run this branch, but that can be tackled later, so I'm approving this PR

PR #421, opened 2021-04-01
Don't use stsynphot, use plain synphot.

mperrin:
Swap to synphot instead of pysynphot for poppy, at basically no loss in functionality we care about. This avoids needing to download many GBs of unrelated reference data. 

The HST calibration spectral dataset is 4.25 GB. That's not acceptable to force as a required download on all users, given the diverse set of use cases for poppy which are in some cases completely unrelated to HST/JWST data. (e.g. optics lab technology work and cubesat optical modeling)

### DIFF
--- a/poppy/utils.py
+++ b/poppy/utils.py
@@ -1465,7 +1465,7 @@ def specFromSpectralType(sptype, return_list=False, catalog=None):
         import os
         cdbs = os.getenv('PYSYN_CDBS')
         if cdbs is None:
-            raise EnvironmentError("Environment variable $PYSYN_CDBS must be defined for stsynphot")
+            raise EnvironmentError("Environment variable $PYSYN_CDBS must be defined for synphot")
         if os.path.exists(os.path.join(os.getenv('PYSYN_CDBS'), 'grid', 'phoenix')):
             catalog = 'phoenix'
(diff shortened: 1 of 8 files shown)


shanosborne (commented):
I'd also look for "stsynphot" in `tox.ini` and in all the docs' `.rst` files

poppy/utils.py
```diff
@@ -1465,7 +1468,7 @@ def specFromSpectralType(sptype, return_list=False, catalog=None):
         import os
         cdbs = os.getenv('PYSYN_CDBS')
         if cdbs is None:
-            raise EnvironmentError("Environment variable $PYSYN_CDBS must be defined for stsynphot")
+            raise EnvironmentError("Environment variable $PYSYN_CDBS must be defined for synphot")
```
shanosborne, line 1471: Does `PYSYN_CDBS` also need to be defined for synphot? Or should this line stay stsynphot? I thought it was stsynphot specific, but I may be wrong

shanosborne (approved):
Besides my 1 comment, this looks good. And I'll let you test that it also works with the synphot-only webbpsf.

mperrin:
That one function - looking up for spectral type to a stellar spectrum - may be the one piece that for now actually does require stsynphot. But that's an optional piece of functionality which many users never use, so I'm OK with that.

Issue #200, opened 2018-08-23
afocal systems break focal length and platescale calculations


mperrin:
<a href="https://github.com/douglase"><img src="https://avatars1.githubusercontent.com/u/1025951?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [douglase](https://github.com/douglase)**
_Tuesday Jan 03, 2017 at 18:02 GMT_
_Originally opened as https://github.com/mperrin/poppy/issues/200_

----

As discovered recently by @neilzim and as [previously](https://github.com/douglase/poppy_example_notebooks/blob/master/Fresnel/Microscope_Example.ipynb) noted but not documented, the current way plate scales are calculated in the fresnel wavefront class depends on the focal length attribute carried through the sequence of propagations by the wavefront class. Thus, it breaks when you have an afocal system where plane-to-plane propagation occurs. 

We need either another way of carrying focal length through a system, or more graceful failure and better documentation.

isa-kain:
Hi @mperrin! I'm running into this bug as well, where the focal length is "forgotten" in collimated space and the plate scale becomes -inf. Has there been any progress on this, or do you have a suggested quick fix?

mperrin:
@isa-kain Thanks for your patience & apologies for not being able to write back sooner. I have not had a chance to look at this yet. 

It might help if you could share some example code that runs into this bug, maybe a simple simulation of an focal simulation like this in a script or notebook. Having a runnable code example would help me get started on looking into this. Cheers.

PR #423, opened 2021-04-05
Added FITSFPMElement class and apply_fits_fpm_fftmft

kian1377:
This PR adds the FITSFPMElement class to poppy_core.py along with the method of applying the FPM called apply_fits_fpm_fftmft to fresnel.py. The method is in the FresnelWavefront class such that the wavefront has the FPM applied to it. There is an additional if statement in the propagate() method of FresnelOpticalSystem to check if the optic type is a FITSFPMElement and if it should be applied using standard multiplication or via the new method.

While working on the SPC modes of the Roman CGI models for POPPY, I still noticed some inconsistencies in the PSFs generated when compared to the results from the PROPER models. The PROPER models would use an FFT/MFT sequence to apply the focal plane masks for the SPC modes, so that sequence is what I have implemented here using POPPY's built in FFT and MFT functionality. 

The sequence does a forward FFT to a virtual pupil plane, it the does an inverse MFT back to the focal plane but with the correct sampling required to apply the FPM, so the FPM amplitude transmission is then applied to the wavefront. A forward MFT is the done to go back to the virtual pupil and an inverse FFT is done to be back at the focal plane with the original wavefront sampling. 

Results from the SPC modes show more accuracy than with my previous models I had implemented via the inwave parameter which I had a made a PR for before. These new results are shown below along with the total flux values in the array extents from both POPPY and PROPER. 

SPC730 mode at 730nm (central wavelength):
image
image

SPC730 mode at 700nm:
image
image

SPC825 mode at 825nm (central wavelength):  
image
image

SPC825 mode at 800nm:  
image
image

I have also verified the display functionality for this type of optic:
image

By making the planetype for the FITSFPMElement be an intermediate planetype, the wavefront at the FPM can also be displayed with extents, meaning the display_intermediates functionality also works: 
image
image

Others involved in this process are @Jashcraf and @douglase . 

Let us know what changes should be made or what should be added for this to become a feature in POPPY.


kian1377:
Hello @mperrin , I just realized that the display functionality is still wrong because although the pixelscale is in units of arcsec/pix and the extents of the figure are correct, the units of the extent still say meters so something would have to be done about that. 

Also, we wanted your input regarding how to write a test for this type of element/method, so if you have any recommendations, please let us know when you can. 

Have a great day.

mperrin:
Hi @kian1377, @douglase, @Jashcraft - First, thanks for this; I can tell a lot of care and cross checking went into this!

Before I get into details of the code too closely I have a couple general questions about the approach:
- To make sure I understand the general picture, the key idea here is to transform via MFT onto a particular fixed sampling (independent of wavelength), apply the existing model of that FPM shape, then transform back? This sounds reasonable, and similar to some of what we do with MFTs in the Fraunhofer coronagraph models, but I want to think through some of the details with sampling. Typically by what factor are you changing the pixel scale from the original image plane to the MFT version? Do we have to worry any about aliasing of frequencies that are not Nyquist sampled in one plane or the other?  (These may not be concerns in practice -- the agreement with PROPER counts for a lot! -- I just have not yet thought through the theoretical side of mixing MFTs into a Fresnel system like this at all.) 
- Can you explain please the reasoning for making the FPM an intermediate plane? Optically it's an image plane of course. It seems perhaps confusing to need to intentionally label an image plane as something other than that. I'd like to think there can be some other way to achieve the necessary steps for propagation that might be more intuitive. If we are getting infinite pixel scales for an image plane, that seems like it is separately a bug that should be investigated and fixed rather than just worked around by setting the plane type differently. Wha
- Can you also explain please the intended inputs for what exactly goes in the FPM file? This shows the FPM shape in units of lamba/D? Or in physical units of meters or microns or similar? I think this could be more clearly documented. 

For the propagation, rather than changing code in the `FresnelOpticalSystem.propagate` function, I think what I will suggest is we add this check in the `FresnelWavefront.__imul__` method. This is what implements the "wavefront *= optic" step. That "looks like" a multiplication but in fact can be a more complicated operation behind the scenes, including the extra Fourier transform operations that you're using here. 

This is a fairly complicated and subtle bit of code, so it's much appreciated! But it's also clearly quite special-purpose to CGI. I want to think about how best to incorporate this into poppy in a generalized manner.  I have some ideas and want to try a couple things out... 

Last question for now. Do you have an example system or a test case for using this? I would want to include in poppy a test function to verify correctness of this, or at least operation-without-crashing. The full CGI model is super impressive but probably overkill for that? It would be great if we could craft a simple small-scale model using this which we can use as the test function. @douglase any ideas for this? I suppose we could put a simple coronagraph into the Fresnel Hubble example for instance.

douglase:
I was thinking the same thing, a simple HST or unobscured Lyot coronagraph would at least tell us if things change substantially going forward. 

There is value to the full CGI and a comparison to a stored PROPER PSF, a lot of errors that @kian1377 chased down were quite small and differences that might be missed in a simpler test case.

kian1377:
Hello @mperrin , thank you for your comments, some of them I can address verbally to you if we can schedule a time to meet via zoom and be able to write more clear documentation for these additions. For now, I can clear up a couple of the concerns you had. 

For one thing, I made the planetype be an intermediate because the pixelscales would go to infinity when I would use an image planetype, but I also think that it does make sense this way because even though this element is at a focal plane, it isn't the actual image plane yet. As for why the pixelscales go to infinity, I think @douglase has a better idea of why thats case and a possible fix to the bug. 

And the intended inputs for the FPM file are simply the file paths for transmission and opd, although when the fpm is applied, the opd is ignored as I haven't yet figured out a way to include it. The FPM file does not need to have any information included in its headers, as I expect the user to provide three parameters as kwargs when calling FITSFPMElement. I suppose functionality can be built to just get those kwarg values from headers as well, but I think that functionality can also be added later. I explain what the kwargs are in the FITSFPMElement code. 

And to answer your first concern last, the idea is definitely to use the MFT to go to the sampling of the FPM array, which is in units of lambda/D (at least for the Roman CGI models). And I did run some tests with different npix and oversample factors for the SPC modes, and the system's would not function properly if npix was ~700, but seemed to work fine when I use npix=800 or above with oversampling of 2. And the oversample factor is taken into account when doing the MFT transform, so when I used 4 or 8, the systems still worked (but were obviously much slower). 

If you are busy and cannot meet anytime soon, I understand, but I think we can clear a lot of things up and share quite a few ideas in a 30min meeting. Just let me know if/when you have time. And I will look into the __imul__ method, it would definitely be more elegant that way.

mperrin:
Thanks! I'm certainly open to meeting by video sometime. Thursday afternoon I have some availability, 2-5 pm EDT so 11am-2pm MST. Let me know if there's some time in that range that would work for you. 

> For one thing, I made the planetype be an intermediate because the pixelscales would go to infinity when I would use an image planetype, but I also think that it does make sense this way because even though this element is at a focal plane, it isn't the actual image plane yet.

Hmm, that's a little different from how the plane type labels were originally intended to work. While the FPM is not the final image plane, it is still an image plane; these planes are optically conjugate to one another. That's distinct from a plane with a finite but nonzero optical conjugate, which is what the intermediate planes are.  Sounds like I should wait and hear what @douglase has to say about the pixel scales. I expect this is related to the issue https://github.com/spacetelescope/poppy/issues/200 reported a couple years ago but not yet prioritized for a fix...

> the opd is ignored as I haven't yet figured out a way to include it. 

This should be straightforward to fix. The necessary steps are to convert from the amplitude transmission and OPD as separate maps to a single 2d complex phasor (via a call to `get_phasor`) and then to apply that to the complex electric field of the wavefront. See in `__imul__` just below where you edited: 

```
        phasor = optic.get_phasor(self)

        if not np.isscalar(phasor) and phasor.size > 1:
            assert self.wavefront.shape == phasor.shape, "Phasor shape {} does not match wavefront shape {}".format(
                phasor.shape, self.wavefront.shape)

        self.wavefront *= phasor
 
```

We'll want to replace your line ` fpm = optic.amplitude` with a call to `get_phasor`, and that should do the trick.

douglase:
yes, it sounds to me like the infinity is #200, since the system  loses knowledge of the plate scale on the sky after going through an afocal relay, but I haven't tested that hypothesis on Kian code yet.

kian1377:
@mperrin , I updated the code such that the FITSFPMElement has opd functionality. I did a simple test by applying a 250nm opd, and the display result is this. 
image

The SPC730 psf now looks like this:
image

Note: I did not use the get_phasor() method because there were additional pixelscale checks that we don't need in the apply_fits_fpm_fftmft() method since the MFTs automatically adjust the pixelscales accordingly to the FPM data.

kian1377:
@mperrin , also, regarding the meeting time, I would be available at 11:30am MST, so 2:30pm EDT.

kian1377:
@mperrin , would you still like to meet tomorrow at 2:30pm EDT. I am still available but no worries if you are too busy to do so.

mperrin:
Apologies but we'll have to aim for another day; my entire afternoon tomorrow is now blocked solid with calls.

kian1377:
No worries, lets try for sometime next week.

kian1377:
@mperrin , would you be available to meet sometime this week? I want to progress on writing tests for this optical element/method as well as address any lingering concerns you may have. Let me know when you can and have a great day.

kian1377:
@mperrin , hello, I just wanted to check in and see what your remaining thoughts are for this PR. 

I have wrote a test case that is reliant on a previously saved FPM fits file and a previously saved result. However, it seems that when that test is being run, the result is not actually accurate to the result I have saved. When I run the test on my machine, there is no numpy assertion error so it passes. Do you have an idea of how I should fix this?

kian1377:
Hello @mperrin , I just revisited this PR because there was still a slight issue with the test I had written for the FITSFPMElement code and method. There were previously slight mismatches in 1 or 2 of the elements of the saved PSF array versus the final PSF generated via propagation in the test function, but now the mismatches do not occur because I increased the relative tolerance threshold from 1e-7 to 1e-6. Given the absolute maximum of the array is 0.0219, I figured this still means the arrays are very close. 

Please let me know your thought on this PR soon because we are mentioning this method in a manuscript we are writing for SPIE Optics & Photonics. 

Regards and have a great day!

poppy/fresnel.py
```diff
@@ -165,6 +165,68 @@ def __init__(self,
         self.K = K
 
 
+class FITSFPMElement(FITSOpticalElement):
```
mperrin: Naming question. The name proposed here, `FITSFPMElement`, doesn't indicate anything about sampling or MTF. Would a different name help provide more clarity for what the purpose of this class is? Also generally we prefer to spell out most acronyms in class names so we could consider spelling out FPM. Perhaps `FixedSamplingImagePlaneElement`? 

This is a relatively minor concern but good API design helps maintain good usability. 

[Optional change]

poppy/fresnel.py
```diff
@@ -951,6 +1018,58 @@ def apply_lens_power(self, optic, ignore_wavefront=False):
...
+        
+        mft = poppy.matrixDFT.MatrixFourierTransform(centering=optic.centering)
+        
+        # self.wavefront = np.fft.fftshift(self.wavefront)
+        # self.wavefront = accel_math.fft_2d(self.wavefront, forward=True, fftshift=True) # do a forward FFT to virtual pupil
+        # self.wavefront = mft.inverse(self.wavefront, nfpmlamD, nfpm) # MFT back to highly sampled focal plane
+        # self.wavefront *= fpm_phasor
+        # self.wavefront = mft.perform(self.wavefront, nfpmlamD, n) # MFT to virtual pupil
+        # self.wavefront = accel_math.fft_2d(self.wavefront, forward=False, fftshift=True) # FFT back to normally-sampled focal plane
+        # self.wavefront = np.fft.ifftshift(self.wavefront)
```
mperrin: This block of commented-out lines should be removed, please.  [Required change]

poppy/fresnel.py
```diff
@@ -165,6 +165,68 @@ def __init__(self,
...
+    when propagating to this plane. 
+    Note: if an image planetype were to be used, the wavefront at this plane may have infinite pixelscales, 
+    making it impossible to display the wavefront with extents.
+
+    The method used to apply this element requires additional information from the user that is not required 
+    for FITSOpticalElements. These additional parameters are listed below. 
+    
+    Parameters not in FITSOpticalElement
+    ----------
+    wavelength_c: astropy.quantity
```
mperrin, line 183: The proposed parameters for this class are inconsistent with much of the rest of poppy, since everywhere else we try to use physical units (arcsec, meters) rather than lambda/D units. I realize there are cases where lambda/D units are helpful, so it's not black or white. But I think this doesn't really follow the principle of least surprise. 

It seems to me that specifying a single `pixelscale` parameter giving the FITS file's pixel scale in `arcsec/pixel` can replace the need for all of the `wavelength_c`, `ep_diam` and `pixelscale_lamD` parameters.  You already calculate that from the provided parameters:
```
pixelscale_lamD * (wavelength_c/ep_diam * 180/np.pi * 3600).value * u.arcsec/u.pix
```
It seems to me that if you simply have the user calculate and provide that parameter it simplifies the interface here and makes it more consistent with the rest of poppy.  Would you consider making that change?  

Alternatively, if you feel strongly that the current API here should stay unchanged, in that case we should: 
1. Make the doc string text state clearly and emphatically that the way sampling is defined for this optic is really quite different than for any other optic. 
2. Consider renaming the input parameters to be more explicit: `center_wavelength`, `entrance_pupil_diameter`, `pixelscale_in_lamD`


[Optional change, but I'd like to at least discuss the pros and cons here and seriously consider revising this, please]

poppy/fresnel.py
```diff
@@ -165,6 +165,68 @@ def __init__(self,
...
+    Parameters not in FITSOpticalElement
+    ----------
+    wavelength_c: astropy.quantity 
+        Central wavelength of the user's system, required in order to 
+        convert the pixelscale to units of lambda/D and scale the 
+        pixelscale of the element based on the wavelength being propagated. 
+    ep_diam: astropy.quantity
+        Entrance pupil diameter of the system, required to convert the 
+        pixelscale to units of lambda/D. 
+    pixelscale_lamD: float
```
mperrin: Independently of the above comment: It would be nice if the sampling parameters can be read from the FITS file header keywords. One of the great things about FITS files is they contain metadata as well as image data. The `FITSOpticalElement` parent class will of course attempt to read the pixel scale from a FITS keyword. This should too, for consistency. 

In particular, when creating some FITS file containing the image data for a FPM mask, isn't it the case that the desired sampling parameters will always be known at that time? So if those values are written into the FITS file header, it ensures the correct values will later be used for that file. This is better than having to separately provide the filename as well as the sampling values in a function call independently; if the values are included in the file themselves there's no way for it to become inconsistent. 

[Optional change]

poppy/tests/test_fresnel.py
```diff
@@ -830,3 +834,59 @@ def test_inwave_fresnel(plot=False):
...
+    hst.add_optic(poppy.CircularAperture(radius=diam.value/2))
+    hst.add_optic(g1)
+    hst.add_optic(g2, distance=d_pri_sec)
+    hst.add_optic(fpm, distance=d_sec_to_focus)
+    hst.add_optic(oap, distance=fl_oap)
+    hst.add_optic(oap, distance=fl_oap)
+    hst.add_optic(poppy.ScalarTransmission(planetype=poppy.poppy_core.PlaneType.intermediate, name='Image'), distance=fl_oap)
+
+    # Create a PSF
+    # Create a PSF
```
mperrin: minor: comment line is duplicated, please delete. [Required change]

mperrin (changes requested):
@kian1377 Apologies for the length of time it's taken until I've had a chance to look at it. It remains the case that I have very little time for anything other than JWST preparations right now. Thanks for your understanding! I started some comments a while ago but didn't have a chance to finish this review until today. 

Overall this is a very nice piece of functionality, and thank you much for contributing it! I have a few questions and suggestions about the function parameters, mostly motivated by thinking about how to make this new class consistent with the rest of poppy so it's as straightforward as possible for users to use. Plus a few places where comments should be cleaned up. 

See below and let me know your thoughts. Thanks again!

kian1377:
@mperrin , sorry I didn't respond last week, but happy Monday!

I did rework the code some such that the pixelscale is supplied as quantity in units of arcsec/pix, but I kept the attribute called pixelscale_lamD because I think it would nice for a user to be able to check that attribute to see what the pixelscale is in lambda/D units when an element is initialized. 

I also changed the code such that the other attributes/keywords can be set through FITS file headers. I implemented these into the header of the FITS file used for the test I wrote for this element and it does pass the test now. 

Also, I liked the name you suggested so I changed the name to FixedSamplingImagePlaneElement. 

Please let me know your thoughts on these changes when you can and have a great day!

poppy/fresnel.py
```diff
@@ -165,6 +165,85 @@ def __init__(self,
...
+        pixelscale value in units of lambda/D.
```
mperrin: Docstring needs an update since the parameter got renamed to `pixelscale`.

Would be nice to include in the doc string here a statement: "If this parameter is left as None, the pixel scale can be read from the FITS header keyword PIXELSCL, if that keyword exists in the provided file."
And a similar statement for the WAVELENC and PUPLDIAM keywords for those parameters too.

[required change] Having clear and complete documentation is a high priority when writing code you'd like other to be able to use :-)

poppy/tests/test_fresnel.py
```diff
@@ -789,44 +793,50 @@ def test_CompoundOpticalSystem_hybrid(npix=128):
                                err_msg="PSFs do not match between equivalent simple and compound/hybrid optical systems")
 
 
-def test_inwave_fresnel(plot=False):
-    '''Verify basic functionality of the inwave kwarg for a basic FresnelOpticalSystem()'''
-    npix = 128
-    oversample = 2
+def test_FixedSamplingImagePlaneElement(display=False):
+    poppy_tests_fpath = os.path.abspath(poppy.__file__)[:-11]+'tests/'
```
mperrin: The `[:-11]` is a bit of a kludge for accessing a directory name here. Better would be something using the os.path functions, for instance `os.path.dirname(os.path.abspath(__file__))`

[optional]

poppy/tests/test_fresnel.py
```diff
@@ -789,44 +793,50 @@ def test_CompoundOpticalSystem_hybrid(npix=128):
                                err_msg="PSFs do not match between equivalent simple and compound/hybrid optical systems")
 
 
-def test_inwave_fresnel(plot=False):
```
mperrin: Something seems to have gone wrong here, possibly in one of the merge commits - I don't think you intend to delete the existing function `test_inwave_fresnel` here, do you? I assume your intent is to add a new additional function `test_FixedSamplingImagePlaneElement`, yes?   

We should not delete the test for the inwave functionality unless deleting that actual functionality, which I don't think is your intent here. 

[required fix, please]

mperrin (changes requested):
Hi @kian1377 This looks very good, thanks for those revisions. I think this is very close and have just two requests at this point: updating a doc string to reflect the changes in your latest commits, and fixing what appears to be the accidental deletion of one of the unit tests. Thanks!

kian1377:
Hello @mperrin , I updated the docstrings for the available keywords, changed the method for obtaining the path to the tests directory to your recommendation, and added back the test_inwave_fresnel function (I did not realize I had accidentally deleted that so sorry if it took you a while to find that issue). 

Let me know if there are any final adjustments required.

mperrin (approved):
All looks good! Thanks much Kian

PR #426, opened 2021-04-07
geometry.py np.arctan divide by zero warning 

### LINKED ISSUE #425: geometry.py np.arctan divide by zero warning
```
anaconda3/envs/astroconda/lib/python3.6/site-packages/poppy/geometry.py:40: RuntimeWarning: divide by zero encountered in true_divide
  return 0.5 * r**2 * (np.arctan(y1/x) - np.arctan(y0/x))
```

> That is a benign and ignorable error encountered in the code for gray-pixel approximations on the aperture edges. Probably affects some pixel right on the x axis, raises the warning, but has no effect on the output PSF. What happens is that y/x goes to infinity if x is 0, but then arctan of infinity is pi/2 so that’s fine. We should fix that to use arctan2 instead of arctan, which doesn’t raise the error

robelgeda:
closes #425 

~~Replaces `np.arctan` -> `np.arctan2` to handle divide by zero warnings.~~

### DIFF
--- a/poppy/geometry.py
+++ b/poppy/geometry.py
@@ -34,10 +34,11 @@ def _arc(x, y0, y1, r):
     is traversed clockwise then the area is negative, otherwise it is
     positive.
     """
-    if accel_math._USE_NUMEXPR:
-        return ne.evaluate("0.5 * r**2 * (arctan2(y1, x) - arctan2(y0, x))")
-    else:
-        return 0.5 * r**2 * (np.arctan2(y1, x) - np.arctan2(y0, x))
+    with np.errstate(divide='ignore'):
+        if accel_math._USE_NUMEXPR:
+            return ne.evaluate("0.5 * r**2 * (arctan(y1/x) - arctan(y0/x))")
+        else:
+            return 0.5 * r**2 * (np.arctan(y1/x) - np.arctan(y0/x))
 
 def _chord(x, y0, y1):
     """
--- a/poppy/geometry.py
+++ b/poppy/geometry.py
@@ -35,9 +35,9 @@ def _arc(x, y0, y1, r):
     positive.
     """
     if accel_math._USE_NUMEXPR:
-        return ne.evaluate("0.5 * r**2 * (arctan(y1/x) - arctan(y0/x))")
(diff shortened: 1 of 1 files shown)


poppy/geometry.py
```diff
@@ -35,9 +35,9 @@ def _arc(x, y0, y1, r):
     positive.
     """
     if accel_math._USE_NUMEXPR:
-        return ne.evaluate("0.5 * r**2 * (arctan(y1/x) - arctan(y0/x))")
+        return ne.evaluate("0.5 * r**2 * (arctan2(y1/x) - arctan2(y0/x))")
```
mperrin: The syntax with arctan2 need two parameters separated by a comma: `arctan2(numerator, denominator)`, instead of a single parameter that's a ratio.

mperrin (changes requested):
Needs syntax change.

poppy/geometry.py
robelgeda: Thank you, I threw together this PR too quickly. I have made the change.

mperrin:
Hi @robelgeda . Following up on this PR from a while ago. it turns out this change is not as simple as we had hoped, because the change from arctan -> arctan2 for some reason causes one of the unit tests to fail. See the Travis logs.  Some detail of the arctan behavior is expected and relied on by the code; I think this is that the arctan return value range is (-π/2, π/2) while the arctan2 range is (-π, π). 

We'll need to try another fix for this. Not quite sure which. It's a small detail but I don't want to leave this hanging indefinitely. I think what might be easiest here is to avoid changing the code but suppress the divide by zero warning. I believe this can be done like this, using a `numpy.errstate` context: 

```
with numpy.errstate(divide='ignore'):
    return 0.5 * r**2 * (np.arctan(y1/x) - np.arctan(y0/x))
```

Can you try that, please? Thanks. 

(non-urgent, but like I said above, I don't want to leave this hanging indefinitely)

robelgeda:
@mperrin Thank you for suggesting that! I made the change. Lets see if things pass this round.

robelgeda:
Looks like that did the trick!

mperrin (approved):

PR #437, opened 2021-05-20
Kolmogorov based propagation through turbulent atmosphere

DaPhil:
Hi @mperrin. If it is still welcome, I added Kolmogorov type phase distortions to simulate propagation through a turbulent atmosphere. I did it a while ago but the repo was shifted.


mperrin:
Hi @DaPhil  - thanks, and yes this is still welcome! Much appreciated for your moving this PR over to this repo. (At one point I had meant to do so myself but never quite figured out how to relocate a PR between repos easily... and then other things became more urgent and I never got back around to it.). Let me take a quick review look at this PR and your other one and hopefully we can have them merged swiftly..

poppy/wfe.py
```diff
@@ -579,7 +579,249 @@ def get_opd(self, wave):
...
+        super(KolmogorovWFE, self).__init__(name=name, **kwargs)
+        
+        self.r0 = r0
+        self.Cn2 = Cn2
+        self.dz = dz.to(u.m)
+        self.inner_scale = inner_scale
+        self.outer_scale = outer_scale
+        self.kind = kind
+    
+    def get_opd(self, wave):
```
mperrin, line 646: Thanks for including that nice set of tests for the Kologorov code. One piece seems to have slipped through the cracks: actually calling the `get_opd` function as part of any of those tests. The Codecov test coverage report (see comments added onto this PR automatically by that) shows that whole function is untested. It would be good to include a test of this, even if a very basic one such as just demonstrating that adding a Kolmogorov phase screen reduces the Strehl of a beam by at least some amount.  

Other than that the test coverage looks excellent. The existing tests cover all the sub components where most of the complexity is; but we should ideally also have a test of the overall top level function to show that puts those components together correctly. Or at least in a way that lets that function be called and return reasonable results by some simple check. Thank you!

mperrin (changes requested):
Looks great overall and thanks again. I have one comment/request about test coverage; see below for what looks like an accidental omission of a test for get_opd, which hopefully will be straightforward and easy to add.

poppy/wfe.py
DaPhil: Hi @mperrin. I just updated my code to include a test of get_opd. However, the checks that run were not all successful. As in my other commit, one test complains about a decrease in coverage of files I never touched! I am not that experienced with version control etc, please advice what I can do to solve that issue.

mperrin: Hi! First, we have seen that bug before from the code coverage system, where it complains about decreased coverage in files never touched. Not sure what's going on with that, but it's definitely a bug, and we generally tend to ignore it when that happens. 

Secondly, on my side right now I don't even see that - it shows all tests passing, including the code coverage over the diff. Transient glitch apparently? While that's not entirely reassuring, it does tend to reinforce our habit of just ignoring that bug when it crops up!

mperrin (approved):
Looks good to me! Tests pass (in CI and locally), and I checked the notebooks runs properly for me too. All seems as it should be. Thanks much @DaPhil

PR #438, opened 2021-05-20
Thermal Blooming Phase Screen

DaPhil:
Hi @mperrin: This is my other addition to poppy for which I hope it is still welcome to be added. It implements a thermal blooming phase screen to be able to calculate how the intensity profile varies due to the heating of propagation medium.


mperrin:
Hi @DaPhil, 

One thing I would like to check with this PR is the sign convention for the OPD. We recently (~ January) revised the sign conventions used in poppy; for consistency with optical modeling tools such as Zemax and Code V, we now have an opposite sign convention than before. See https://poppy-optics.readthedocs.io/en/latest/sign_conventions_for_coordinates_and_phase.html#Sign-Conventions-for-Wavefront-Error-and-Phase for details. 

I expect this is relevant for the thermal blooming OPD so I wanted to check about this. Can you confirm whether this PR is consistent with the sign conventions as stated on that page, or should we flip the sign on the OPD in this?

DaPhil:
Hi @mperrin. In my derivations for the implementation of all the equations I used the old sign convention exp(+i(kz - wt)). I need to check wether this has an implication on how to create the phase screen. I think not, since I tested my code against literature. Check my notebook on thermal blooming [here](https://github.com/DaPhil/poppy/blob/thermal_blooming/notebooks/Thermal%20Blooming%20Demo.ipynb) and you see that the code reproduces the papers result with identical input. But as I said, I will check my notes (from a couple of years ago) to see if I can spot something that would need modification. I will also check this for the Kolmogorov phase screen.

DaPhil:
Hi @mperrin. As far as I can see, you did not change how phases are applied. In optics.py -> AnalyticalOpticalElement -> get_phasor, the phase was applied via exp(+i k OPD) in the past and it is still the case. Hence, no change in the phase required (neither for thermal blooming nor for Kolmogorov phase screens). Now my question is, what do I have to do to make this pull request pass everything? I probably have to make more elaborated tests? Is there anything else?

mperrin:
Hi @DaPhil, 

Thanks again for your patience while other priorities have had my attention elsewhere... I'm finally getting back to this. 

I merged master to this and de-conflicted edits in wfe.py. Now waiting to see if this tests run. First try didn't work but that was my error from an imperfect merged; Fixed that now and trying again. 

> as far as I can see, you did not change how phases are applied. In optics.py -> AnalyticalOpticalElement -> get_phasor, the phase was applied via exp(+i k OPD) in the past and it is still the case. 

Yes and no. We didn't change the how phases are applied, however we did change the sign of the FFT in the Fourier propagation step, which results in changing how the phase affects the beam as it propagates. The updated sign convention is now more strictly consistent with that of e.g. Zemax or Code V, though it differs from that in Goodman's Fourier optics. Can't be consistent with everyone... To be specific the sign convention is now such that positive WFE represents an advanced wavefront (an OPD region with positive OPD leads compared to the surface with zero WFE; see https://poppy-optics.readthedocs.io/en/latest/sign_conventions_for_coordinates_and_phase.html#Sign-Conventions-for-Wavefront-Error-and-Phase). If your thermal blooming phase screen is consistent with that then it's good to go.

mperrin:
Well this is curious - the tests _mostly_ pass , except the test matrix case using the accelerated math library numexpr for some reason doesn't pass: 

```
        rho = phase_screen.rho_dot_FT(wf)
        assert(rho.shape[0] == 1024)
        assert(rho.shape[1] == 1024)
>       assert(np.round(np.max(rho), 6) == np.round((1.936370481716403-0.005940728873190856j), 6))
E       assert (1.93637+0.005941j) == (1.93637-0.005941j)
```

See https://github.com/spacetelescope/poppy/pull/438/checks?check_run_id=3090030722

Not sure what's going on with this - it may take some more digging in.

DaPhil:
Hi @mperrin. First, the phase screen is consistent with all your definitions. Second, the test that fails was supposed to be taken out. The function is implicitly tested by the non-isobaric phase screen test. Hence I do not think it is even necessary. The problem is a sign flip of the near-zero imaginary part. When I wrote the test, it was based on an x-direction velocity, but here the test uses a y-direction velocity because I added a lot of tests (and forgot to check this one thoroughly). So I see 2 options: either leave the test out completely or adapt it to pass. Both are easy to do. I prefer deleting this part of the test. What do you prefer?

mperrin:
I’m happy to follow your suggestion and delete that part of the test!

mperrin:
@DaPhil If you'd like I can go ahead and delete the unnecessary test myself, or if you prefer you can do it. Either way is fine with me - just let me know if you'd like me to do it myself.

DaPhil:
@mperrin Please go ahead and delete it.

mperrin:
Looks good - all tests pass now! (The code test coverage check would like more unit test coverage but the un-tested code is mostly error condition checking and else cases, not a big deal and certainly not something to hold up this PR any further for). 

Thanks for adding this @DaPhil. It's way outside my own area of expertise in multiple ways (most of the optics I'm modeling are far above the atmosphere, and the idea of light power strong enough to cause thermal deformations is also not exactly typical of astronomical signals :-) So, thanks for contributing your expertise to broaden out the kinds of optical physics we can model in poppy! Cheers.

PR #441, opened 2021-06-22
Improved Test Gaussian Beam curvature near beam waist 

zufus:
The Fresnel test suite contains a test-case to assess the ability to calculate the curvature and spreading near the beam waist.
The actual test uses the methods spot_radius() and r_c() to evaluate the spreading and the wavefront radius.

However, the spot_radius() and the r_c() methods use exactly the same analytical formula used to validate the results, rather than the numerical propagated wavefront.

I believe it is more "reliable" to define a Gaussian beam, propagate it, evaluate the intensity/phase distribution (on the propagated field) and then compare the results with the analytical model.

I hope this helps! (I am quite new to poppy, but I find it really exciting!)

### DIFF
--- a/poppy/tests/test_fresnel.py
+++ b/poppy/tests/test_fresnel.py
@@ -1,6 +1,3 @@
-import pytest
-
-import poppy
 from .. import poppy_core
 from .. import optics
 from .. import misc
@@ -8,11 +5,13 @@ from .. import fresnel
 from .. import utils
 from poppy.poppy_core import _log, PlaneType
 
+
 import astropy.units as u
 import matplotlib.pyplot as plt
 import numpy as np
 from .. import fwcentroid
 from scipy.ndimage import zoom,shift
+from scipy.optimize import curve_fit
 
 def test_GaussianBeamParams():
     """Confirm that gaussian beam parameters agree with expectations"""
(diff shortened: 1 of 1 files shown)


poppy/tests/test_fresnel.py
```diff
@@ -169,7 +265,7 @@ def test_Circular_Aperture_PTP_long(display=False, npix=512, display_proper=Fals
 
     # also let's test that the output is centered on the array as expected.
     # the peak pixel should be at the coordinates (0,0)
-    assert inten[((y==0) & (x==0))] == inten.max()
+    assert inten[np.where((y==0) & (x==0))] == inten.max()
```
mperrin, line 268: `np.where` is not needed, and is slower in performance than indexing with boolean arrays. Please undo this change.   Or perhaps this is also an issue from not having merged the latest develop into this PR, or some similar git conflict issue?

poppy/tests/test_fresnel.py
```diff
@@ -185,122 +281,43 @@ def test_Circular_Aperture_PTP_long(display=False, npix=512, display_proper=Fals
...
 
-try:
-    from skimage.registration import phase_cross_correlation
-    HAS_SKIMAGE = True
-except ImportError:
-    HAS_SKIMAGE = False
-
-@pytest.mark.skipif(not HAS_SKIMAGE, reason='This test requires having scikit-image installed.')
-def test_Circular_Aperture_PTP_short(display=False, npix=512, oversample=4, include_wfe=True, display_proper=False):
+def test_Circular_Aperture_PTP_short(display=False, npix=512, display_proper=False):
```
mperrin, line 284: There seem to be a lot of changes and missing/deleted code that aren't related to the main point of this PR. I expect this is likely a git issue from making this branch from an outdated branch rather than the most recent develop.  

This will need to be cleaned up before we could consider merging this PR, please. Can you make sure the most recent develop is merged into your PR, please?

poppy/tests/test_fresnel.py
```diff
@@ -26,6 +25,103 @@ def test_GaussianBeamParams():
...
+        print ("Propagating to {:0.2e}".format(zi))
+        y, x = wf.coordinates()
+        pw = [1, 0, (w0 * np.sqrt(1.0 + (zi / z_rayleigh) ** 2))]    
+        popt_w, pcov_w = curve_fit(gauss, x[wf.intensity.shape[1]//2,:], wf.intensity[wf.intensity.shape[1]//2,:], p0 = pw)
+        
+        print(pw[2], popt_w[2])
+        
+        try:
+            assert np.allclose(wf.intensity[wf.intensity.shape[1]//2,:],
+                               gauss(x[wf.intensity.shape[1]//2,:], popt_w[0], popt_w[1], popt_w[2]))
```
mperrin, line 73: The `assert` statement should have a second argument which is a string error message that will be displayed if the assertion fails.

poppy/tests/test_fresnel.py
```diff
@@ -26,6 +25,103 @@ def test_GaussianBeamParams():
...
+        wf.propagate_fresnel(zi)
+        
+        print ("Propagating to {:0.2e}".format(zi))
+        y, x = wf.coordinates()
+        pw = [1, 0, (w0 * np.sqrt(1.0 + (zi / z_rayleigh) ** 2))]    
+        popt_w, pcov_w = curve_fit(gauss, x[wf.intensity.shape[1]//2,:], wf.intensity[wf.intensity.shape[1]//2,:], p0 = pw)
+        
+        print(pw[2], popt_w[2])
+        
+        try:
```
mperrin, line 71: This pattern of `try/except` is not correct in a unit test: a test assertion failure _should_ raise an exception so the test fails, but this pattern would catch the exception and suppress it. The `try/except` should be removed. The optional plotting code should be wrapped in an `if plot:` statement instead.

poppy/tests/test_fresnel.py
```diff
@@ -26,6 +25,103 @@ def test_GaussianBeamParams():
...
+            plt.savefig("intensity_error_at_{:0.2}.png".format(zi))
+            plt.close()
+            
+        pr = [1, (zi*(1 + (z_rayleigh/zi)**2)), 0]    
+        popt_r, pcov_r = curve_fit(gauss, x[wf.phase.shape[1]//2,:], np.unwrap(wf.phase[wf.phase.shape[1]//2,:]), p0 = pr)
+        
+        print(pr[1], popt_r[1])
+        
+        
+        try:
```
mperrin, line 89: Same comments as above.

poppy/tests/test_fresnel.py
```diff
@@ -26,6 +25,103 @@ def test_GaussianBeamParams():
     assert(np.round(gw.z_r.value,9) == np.round(0.042688650889351865,9))
     # FIXME MP: where do the above values come from?
 
+def gauss(x, *p):
```
mperrin, line 28: Function needs doc string, at least a simple one-line one.

poppy/tests/test_fresnel.py
```diff
@@ -26,6 +25,103 @@ def test_GaussianBeamParams():
...
+    A, mu, sigma = p
+    return A*np.exp(-2*(x-mu)**2/(sigma**2))
+
+def radius(x, *p):
```
mperrin, line 32: Same: Function needs doc string, at least a simple one-line one.

poppy/tests/test_fresnel.py
```diff
@@ -26,6 +25,103 @@ def test_GaussianBeamParams():
...
+    k, r, c = p
+    return k + k*(x**2)/(2*r) + c
+
+    
+def test_Gaussian_Beam_zufus(npoints=5, plot = False):
```
mperrin, line 37: Same: Function needs doc string, at least a simple one-line one.

Also let's please rename this to something more descriptive of what it does rather than with your username in it (which is not particularly informative about what the test is testing!)

mperrin (changes requested):
Hi @zufus, Thanks very much for this contribution. After reviewing the code I have some suggestions and a few requests for changes; see below.  

In particular the biggest issue is it looks like this PR is out of sync with the current develop branch and seems to be missing some edits to these files from at least PR #402 and perhaps other PRs; that'll need to be fixed before this could be merged.  We should make sure that the diff display in Github shows only the particular changes that are actually intended. 

Thanks again!

mperrin:
@zufus  - this is a follow up on a rather old PR, which you submitted in this repo about 2 years ago. At the time, I reviewed the PR and requested some changes. We didn't hear back from you then, and I expect this is all out of your recent memory in any case.  I'm trying to decided if we should let this PR go (i.e. close it and do not merge), or if you might have some bandwidth to return to it and help us get it finished and merged.  If you don't currently have interest or availability to help with this, that's fine, please just let me know and we can proceed without you. 

Once again, I appreciate your contribution of this PR back then, and I apologize for the lengthy delay. We were all rather busy in 2021-22 with preparations for and then launch and commissioning of JWST... Thanks again.

mperrin:
No response on this. Going to mark this stale PR as closed without further action.

Issue #436, opened 2021-05-14
Undefined pupil path variable


JarronL:
Minor issue (that I haven't run into, but Pylance in VS Studio Code discovered it while I had the file opened): full_pupil_path has not been defined at the point in which it is called when raising the file not found error.

https://github.com/spacetelescope/poppy/blob/ae26bee53381e38834ccbfd63350d5d938c9c6e8/poppy/instrument.py#L570

Instead, this should simply read:
```python
raise IOError("File not found: " + self.pupil)
```

PR #444, opened 2021-07-16
fix minor syntax error for #436

mperrin:
See #436


mperrin:
Merging without review; this is a one-line fix for a small syntax error that never crops up in practice.

Issue #409, opened 2021-03-18
PEP8 utils


mperrin:
@shanosborne noted on #394 
> we could take this opportunity to replace utils.removePadding with utils.remove_padding to make it more Pythonic. But that may also fall into a backward compatibility hole, so there's no reason we have too.

We should PEP8 all the utils functions, yes.

mperrin:
Closed by #446

PR #446, opened 2021-07-16
PEP8 cleanup in utils.py

mperrin:
Addresses #409:
 - Replace a few function names in utils.py with PEP8-compatible ones
 - PEP8 cleanup generally throughout utils.py
 - Remove a deprecated function `measure_strehl` which was long ago deprecated here and moved to WebbPSF.


mperrin:
FYI @shanosborne : Minor PR here with some PEP8ification. Should be simple and quick to review; thanks.

shanosborne (approved):
Looks good!

PR #447, opened 2021-07-19
allow shifts of optics to be specified with astropy Quantities. 

### LINKED ISSUE #387: allow astropy.units in shift_x or shift_y parameters to OpticalElement
Moved from #378:

Hmm, here's another issue with units that should work but doesn't:
```
ap1 = poppy.RectangleAperture(width=50*u.micron, height=300*u.micron, shift_x = 100*u.micron)
ap1.display()
```
You can't specify shift_x or shift_y with units, just as a bare float with implicit units. This too should be cleaned up and made consistent & unsurprising.

mperrin:
Fixes #387, allowing use of Quantities when shifting optics. This works now: 
```
ap1 = poppy.RectangleAperture(width=50*u.micron, height=300*u.micron, shift_x = 100*u.micron)
ap1.display()
```

### DIFF
--- a/poppy/optics.py
+++ b/poppy/optics.py
@@ -300,8 +300,11 @@ class AnalyticOpticalElement(OpticalElement):
         "shift_x", "shift_y", "rotation", "inclination_x", "inclination_y"
         If any of them are present, then the coordinates are modified accordingly.
 
-        Shifts are given in meters for pupil optics and arcseconds for image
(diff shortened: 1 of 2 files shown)


mperrin:
@shanosborne FYI another pretty small PR closing an open issue about units

shanosborne (approved):

PR #448, opened 2021-07-20
improve unit tests for Kolmogorov WFE

mperrin:
Improvements to the unit tests for the Kolmogorov WFE class, to fix two practical issues. (This isn't a high priority right now in itself, but I am trying to make the whole test suite more robust): 

1. use a fixed seed for random number reproducibility. This fixes an issue in which, intermittently, the random number generator apparently produced a WFE instance which didn't quite meet the requirements. This resulted in rare failures for the CI tests. Making this a fixed seed for the RandomState ensures test repeatability. 
2. 10x reduced test runtime for an ensemble statistics test, `test_KolmogorovWFE_correlation`, which previously took 40% of the total runtime for the entire poppy test suite (~ 100 s out of 250 s total!). This is disproportionate for a single test of one part of functionality. This was reduced to ~10 s by dropping to a smaller number of wavefronts in the ensemble (800 vs 2000 before) and reducing the size of the WFE in spatial sampling (32 pixels vs 64 pixels before). Using smaller values than those resulted in the test not passing, presumably because of averaging over an insufficient number of statistical instances. 
3. Also, ensure the `test_KolmogorovWFE.test_get_opd` function actually gets ran and tested. Previously it was defined but not actually called. And then make that test pass in all instances of the test matrix.

### DIFF
--- a/poppy/tests/test_wfe.py
+++ b/poppy/tests/test_wfe.py
@@ -282,11 +282,14 @@ def test_PowerSpectrumWFE(plot=False):
 def test_KolmogorovWFE():
     CN2 = 1e-14*u.m**(-2/3)
     DZ = 50.0*u.m
+
+    seed = 12345678  # Use fixed seed to ensure reproducible test behavior.
+                     # This particular value is arbitrary; any 32-bit int should work here.
     
     def test_KolmogorovWFE_stats():
         # verify statistics of random numbers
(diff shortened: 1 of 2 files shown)


mperrin:
**Without this PR:** 

Local test runtime: ~ 250 seconds
Github actions CI test runtime:  ~ 9-10 minutes

**With this PR:**

Local test runtime: ~ 170 seconds
Github actions CI test runtime:  ~ 5-6 minutes

PR #453, opened 2021-08-05
Don't mutate "global" ``self.options``

jamienoss:
This change is based purely on the existing inline comment:

```python
 #... Don't just modify the global self.options
 # structure since that would pollute it with temporary
 # state as well as persistent state.
```


mperrin (approved):
Thanks! Good catch, much appreciated.

PR #454, opened 2021-08-17
Update PowerSpectrumWFE demo notebook

jlumbres:
@douglase found some bugs with my original demo notebook, which was not updated after PowerSpectrumWFE was finalized and merged. The notebook is now up to date with some other small edits too.


mperrin (approved):
I checked out the notebook and verified that I could run all cells without error. Thanks @jlumbres and @douglase

PR #456, opened 2021-08-30
Fix failures in CI

shanosborne:
This PR fixes the failing latest dependencies check that can be seen in #455. The failure seems to come from github actions now looking for the requirements.txt file in the tmp directory rather than the repository. I don't know where this change came from, but I fixed it by copying over the requirements file to the tmp directory first, and now the CI finds the file.


mperrin (approved):
Looks good. Thanks for figuring this out!

PR #460, opened 2021-09-29
Allow for non-integer oversampling in Fresnel propagation

kian1377:
This PR contains changes in just a couple of lines of code that allow for non-integer oversampling to be utilized in Fresnel propagation. One line is in fresnel.py, and it is a simple change to the log message for what the array size has been padded to. The change allows for floats to be printed in the message. The other line is in utils.py in the pad_to_oversample function. It determines the array size that the padded array should be initialized to by taking the values of npix and oversample and then using np.round and int to form the array size value that can be given to np.zeros. 

On my personal machine, the code ran fine and passed all the expected tests. The built in Fresnel propagator seems to produce accurate results and calculate the correct pixelscales for the FresnelWavefronts generated. This has been found empirically by comparing wavefronts generated by poppy and wavefronts of the same systems generated by proper. 

@douglase was also involved in this PR.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -333,7 +333,7 @@ class FresnelWavefront(BaseWavefront):
         if self.oversample > 1 and not self.ispadded:  # add padding for oversampling, if necessary
             self.wavefront = utils.pad_to_oversample(self.wavefront, self.oversample)
             self.ispadded = True
-            logmsg = "Padded WF array for oversampling by {0:d}, to {1}.".format(
+            logmsg = "Padded WF array for oversampling by {0:.3f}, to {1}.".format(
                 self.oversample,
                 self.wavefront.shape
             )
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -1261,7 +1261,7 @@ class FresnelOpticalSystem(BaseOpticalSystem):
             A wavefront appropriate for passing through this optical system.
 
         """
-        oversample = int(np.round(1 / self.beam_ratio))
+        oversample = 1 / self.beam_ratio
         if inwave is None:
             inwave = FresnelWavefront(self.pupil_diameter / 2, wavelength=wavelength,
                                       npix=self.npix, oversample=oversample)
--- a/poppy/tests/test_fresnel.py
+++ b/poppy/tests/test_fresnel.py
@@ -884,3 +884,64 @@ def test_FixedSamplingImagePlaneElement(display=False):
                                err_msg="PSF pixelscale of this test does not match the saved result.", verbose=True)
     
 
+def test_fresnel_noninteger_oversampling(display_intermediates=False):
+    '''Test for noninteger oversampling for basic FresnelOpticalSystem() using HST example system'''
+    lambda_m = 0.5e-6 * u.m
+    diam = 2.4 * u.m
+    fl_pri = 5.52085 * u.m
+    d_pri_sec = 4.907028205 * u.m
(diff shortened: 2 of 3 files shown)


mperrin:
@kian1377 and @douglase thanks much for this. After a brief review this looks good. A couple suggestions/questions: 

- Should we also make a similar change in the Fraunhofer code, so the API for oversampling is consistent between the two? To me it seems best to support non-integer oversampling in both if we support it in either. 
- What about a test case? The current test suite only includes integer oversampling, for either Fresnel or Fraunhofer modes. It would be good to include a basic test that specifies a non-integer oversampling and tests that the resulting image pixel scale is as intended.

kian1377:
@mperrin Sounds good. I haven't used a Fraunhofer system in a while so I will see if I can make a change for that. As for the test case, I will see if I can just alter one of the existing tests slightly and just use a different oversampling factor.

mperrin:
For Fraunhofer I think all that's needed is a similar change in the string formatting. Search in poppy_core.py for the same string "Padded WF array for oversampling [etc]" and it should be straighforward.

kian1377:
@mperrin , So I have looked into the suggestions you had made and to start off with the Fraunhofer non-integer oversampling, it seems that no change is actually required in order to get poppy to function. I ran one of my systems with an npix of 1000 and an oversampling factor of 2.048 and I got 2048x2048 arrays at the expected planes. 

For the PSF of the Fraunhofer system, I was using osys.add_detector(pixelscale=0.01, fov_arcsec=5) and these are the results I got. So it seems the pixelscale is adjusting correctly in order to acquire the correct FOV for the detector.

PSF with oversampling 2.048.
image

PSF with oversampling 2.
image

As for the test that would check the functionality for a Fresnel system, I found that this feature would not be functional with planes that are of the type PlaneType.image. The pixelscales for these planes are calculated in angular units, and after some testing with a simple HST example, it seems the image plane pixelscale would not be adjusted based on the non-integer oversampling. 

However, this does seem to be functional for PlaneType.intermediate planes since these planes contain units in m/pix. But, there is an additional caveat that I have found, which is that the pixelscale adjustments only seem to be made when a user utilizes the inwave functionality with the oversample factor they choose. Below, I have included some examples with the testing I did with the HST Fresnel model. 

These PSFs were created without using an input wavefront. 
image

These PSFs were created with using an input wavefront. 
image

In the example with an input wavefront being used, the shape and pixelscale at the focal plane are different because the 2.5 oversample factor makes the pixelscale at the PSF become smaller (higher resolution). 

I have also included many more examples showing the FPM wavefront for the Roman SPC730 mode, which I initialize using a FixedSamplingImagePlaneElement. The PROPER models for this mode utilize an npix of 1000 and an oversample factor of 2.048 and produce a pixelscale almost exactly matching the one POPPY calculated for the same parameters shown. 

FPM with oversampling 1.8.
image

FPM with oversampling 2.
image

FPM with oversampling 2.048.
image

I am still looking into how to fix the little caveat I mentioned so I will give you an update soon.

kian1377:
I added a test which runs through 3 different HST models. It compares the PSFs for oversampling values of 2 and 2.0 to make sure those come out the same and it checks the pixelscale value for an oversampling value of 2.5. The value the pixelscale is being compared to is hardcoded into the test and it is in units of arcsec/pix. This is because I realized I was wrong about the PlaneType.image not functioning correctly for noninteger oversampling values and when I retested this functionality, I found it works as expected.

mperrin (approved):
Very nice, thank you @kian1377. All looks good to me.

PR #472, opened 2021-12-03
update minimum vers of dependencies

mperrin:
Update numpy, scipy, matplotlib minimum dependencies to versions as of mid 2020.

Like PR https://github.com/spacetelescope/poppy/pull/415, but it's been 9 months since that so we might as well freshen up again prior to 1.0 release.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,9 +19,9 @@ max-line-length = 120
 zip_safe = False
 packages = find:
 install_requires =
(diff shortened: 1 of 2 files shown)


mperrin:
Given short timescales and the need to get a 1.0 release out sooner rather than later, I'm going to just go ahead and merge this; it doesn't really need a review, and all tests pass.

PR #476, opened 2021-12-07
Fix documentation formatting issues, Release 1.0.1...

mperrin:


shanosborne:
@mperrin I tested the docs spacing fix both with `make html` and `sphinx-build docs _build`, so hopefully it should work for you too. Let me know if you see any problems with the HTML pages

PR #487, opened 2022-01-27
Fix failing rotation test from numpy 1.22

shanosborne:
This PR fixes a test that is failing for newer versions of numpy (>= 1.22). In the newest numpy, they have changes in the umath functions to speed them up, but as part of that change, they have also decreased precision. This is not a problem for most of our tests, because the decreased precision is still within the tolerance, but this test works differently.

The test_rotations() test rotates the optical element by -45 and 45 degrees and compares the results. Inside this code, the AnalyticOpticalElement objects use np.sin() and np.cos() to do this rotation, and then does a comparison on the output values and uses that to set the array to 1s or 0s. So that slight precision change in the x and y values leads to a number being miss-set as 1 or 0. So it is a tolerance problem.

We fixed this by changing the test's sampling to 255 in order to avoid the precision being a problem.

### DIFF
--- a/poppy/optics.py
+++ b/poppy/optics.py
@@ -329,8 +329,8 @@ class AnalyticOpticalElement(OpticalElement):
             angle = np.deg2rad(self.rotation)
             xp = np.cos(angle) * x + np.sin(angle) * y
             yp = -np.sin(angle) * x + np.cos(angle) * y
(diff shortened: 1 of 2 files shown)


mperrin (approved):
Looks good. Discussed & collaborated via Slack to come up with this fix.

Issue #483, opened 2022-01-16
`zernike` module raises `TypeError` in `fractorial` for Python 3.10


vandalt:
Hello,

When doing the [WebbPSF quickstart tutorial](https://nbviewer.org/github/spacetelescope/webbpsf/blob/develop/notebooks/WebbPSF_tutorial.ipynb) on Python 3.10, I got a `TypeError` because of the `factorial` call in `zernike.R()`. I think this is because calling `factorial` with a float is [deprecated since 3.9](https://docs.python.org/3/library/math.html#math.factorial).

I think the fix is to just change the divisions to integer divisions, so I can send a PR with that change if you want (I tried it and all unit tests pass again when changing this).

<details>
  <summary>Code to reproduce:</summary>
  <p>

  ```python
  from poppy.zernike import zernike_basis_faster
  a = zernike_basis_faster()
  ```
  </p>

  (Running the unit tests on `develop` with Python 3.10 also reproduces)

</details>

Thank you !

vandalt:
Fixed by #493

PR #493, opened 2022-02-15
Correct zernike.py factorial() calls to ensure internal values are integers

AarynnCarter:
Fix typing of values within factorial() calls to ensure they are integers. Prevents the following error with python 3.10 (also mentioned in issue #483):

```
  File "/Users/acarter/anaconda3/envs/coro_pipeline/lib/python3.10/site-packages/poppy/zernike.py", line 424, in cached_R
    (factorial(k) * factorial((n + m) / 2. - k) * factorial((n - m) / 2. - k)))
TypeError: 'float' object cannot be interpreted as an integer
```


mperrin (approved):
Looks good. Thanks @AarynnCarter!

PR #496, opened 2022-03-11
Add jupyter as a dependency for docbuild with nbsphinx

shanosborne:
Sometime in the last 10 days, the doc building test started failing here and on webbpsf with `Extension error: Could not import extension nbsphinx (exception: No module named 'ipython_genutils')`.

This is fixed by adding `jupyter` as a dependency for the testing suite

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -40,6 +40,7 @@ deps=
     sphinx-issues
     sphinx-astropy
     nbsphinx
+    jupyter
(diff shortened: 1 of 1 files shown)


shanosborne:
It seems the nbsphinx failures fixed themselves since I opened this PR. Not sure I understand why, but let's not add an extra dependency.

PR #498, opened 2022-03-19
Fix Failing CI

shanosborne:
The Astropy Development CI test was failing because there was and issue with our installation command, so I've updated that in the tox.ini file.


mperrin (approved):
👍

Issue #452, opened 2021-08-04
StatisticalPSDWFE divide by zero


philhinzatucsc:
I have a newbie question/ bug.  I have just started using poppy to estimate wavefront error effects on the PSF for Keck/SCALES.  To do so, I have a created a FresnelOptical System model and then use StatisticalPSDWFE to create an error for one of the optics.  This results in a warning:

 . . ./poppy/wfe.py:349: RuntimeWarning: divide by zero encountered in power
  psd = np.power(rho, -self.index)   # generate power-law PSD

I see why it's complaining. r^-3 is infinite when r=0 (center of the array).  So I went to  wfe.py and added a really crude hack:
    rho[rho == 0] = 0.00001   #get rid of infinity.  

Now this works fine and I can happily continue modeling.  Is this a real bug, or am I missing something about how to properly use StatisticalPSDWFE?

mperrin:
@jlumbres could you please take a look at this issue and provide your take on this divide by zero issue with StatisticalPSDWFE? Thanks much..

jlumbres:
Oh, hi @philhinzatucsc! 

This issue happened with me when I wrote PowerSpectrumWFE. I did a similar solution, too. I overwrote the one pixel where spatial frequency k=0 to be 1*dk (spatial frequency spacing), did my math, then I overwrote the power at that same pixel location to be 0 after the math was done. I set the power to 0 because it was an arbitrary choice on my end. You can see my code [here](https://github.com/spacetelescope/poppy/blob/c189bceab789ee8ca22f6ffa54272bcfc095a356/poppy/wfe.py#L531).

philhinzatucsc:
Ok, thanks for the confirmation.  I am new enough to github and poppy that I am not sure what the next step is.  For my own work, my internal hack works fine, but presumably I or you should be fixing this for others? I am not sure how to submit a bug fix.

PR #499, opened 2022-03-28
Addition of CuPy as an Accelerated Computing Option

kian1377:
This is a bit more of an extensive PR as it seeks to add CuPy as an accelerated math option for many computations including FFTs, MFTs, and exponentials for propagation with the OpticalSystem or FresnelOpticalSystem classes. To start with, CuPy is a package for GPU accelerated computing. While POPPY has GPU computing options available with PyOpenCL and with pyculib (which has been deprecated by Numba in favor of CuPy), this implementation seeks to perform all calculations on the GPU until the end of propagation. This reduces time for calculations as arrays no longer need to be transferred between GPU memory and standard memory when performing different calculations. 

CuPy has been designed to be very similar to Numpy and has many of the same features requiring the same syntax to use. An example is numpy.fft.fft2, which with CuPy is cupy.fft.fft2. The catch is that CuPy functions can not be used on Numpy arrays since Numpy arrays are stored on standard memory. As such, the implementation strategy was to use the statement “import cupy as np” when POPPY’s Config has been set to use CuPy. This makes performing all calculations on GPU more seamless as wavefront arrays for Wavefront or FresnelWavefront objects along with optical element phasors are automatically calculated as CuPy arrays. 

In addition to using CuPy for basic computations of FFTs and exponentials, CuPy also enables the use of many SciPy functions through its cupyx functions. The cupyx functions can be imported much like many scipy functions with a statement such as “import cupyx.scipy.ndimage as ndimage”. So many functions for computing array rotations or interpolations are also imported through cupyx or scipy based on POPPY’s Config setup. The same method is also used to compute Bessel functions. 

Currently, many optics have been tested for use with CuPy. A table with these optics is provided below with some comments regarding the functionality of some optics. 
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-a6e38066-7fff-75fd-8214-a1567fec4732"><div dir="ltr" style="margin-left:0pt;" align="left">

Optical Element Class | Compatible with CuPy | Comments
-- | -- | --
CircularAperture | Yes |  
MultiCircularAperture | Yes |  
SquareAperture | Yes |  
RectangularAperture | Yes |  
HexagonAperture | Yes |  
MultiHexagonAperture | Yes |  
NgonAperture | Yes |  
SecondaryObscuration | Yes |  
AsymmetricSecondaryObscuration | Yes |  
CompoundAnalyticOptic | Yes |  
ThinLens | Yes |  
GaussianAperture | Yes |  
KnifeEdge | Yes |  
SquareFieldStop | Yes | These image plane elements have only been tested in Fraunhofer systems since it is difficult to use these elements in Fresnel systems anyway. This is because of the issue in converting to angular units in a Fresnel system.
RectagularFieldStop | Yes
AnnularFieldStop | Yes
HexagonFieldStop | Yes
CircularOcculter | Yes
BarOcculter | Yes
BandLimitedCoronagraph | Yes
IdealFQPM | Yes
ScalarTransmission | Yes |  
InverseTransmission | Yes |  
ScalarOpticalPathDifference | Yes |  
ZernikeWFE | Yes |  
KolmogorovWFE | Yes | Does not pass test when using Tatarski power spectrum.
SineWaveWFE | Yes |  
StatisticalPSDWFE | Yes | Had to implement hack found in Issue #452, pre-existing bug not related to CuPyNote: slight differences in numpy vs cupy random generators so does not produce the exact same result with the same seed value.
PowerSpectrumWFE | Yes | Had to assume opd is calculated in nanometers and then rescale it to meters, there should be a better solution to this. Note: slight differences in numpy vs cupy random generators so does not produce exact same result
ContinuousDeformableMirror | Yes | Influence function must be provided. When using .set_surface(), you should still provide a numpy.ndarray instead of a cupy.ndarray.
HexSegmentedDeformableMirror | Yes | These segmented optics are functional but when converted to an ArrayOpticalElement with fixed_sampling_optic(), the OPD is only 0. Still not 100% sure why but it does not seem to be an issue with CuPy because get_opd() functions as expected.
CircularSegmentedDeformableMirror | Yes
ArrayOpticalElement | Yes |  
FITSOpticalElement | Yes |  
FixedSamplingImagePlaneElement | Yes |  

</div></b>

All tests in the test suite for POPPY were also run. Currently, there is only an issue with the KolmogorovWFE test not passing due to a units issue in an exponential. This only comes up when using a Tatarski power spectrum. The tests have only been run with standard CPU computations to make sure POPPY is not running into critical errors because of the addition of CuPy even if a user isn’t using the CuPy feature. 

Computation comparisons have been performed to illustrate the benefit of this accelerated computing feature. Below are comparisons of the times required for a PSF to be calculated for varying array sizes using the MKL FFT option versus the CuPy calculations. The optical systems tested had 5 different surfaces/optics. The system used for these comparisons was the University of Arizona’s HPC Puma nodes. The node utilized 32 AMD EPYC 7642 CPUs and the NVIDIA Tesla V100S GPU. 

Propagation Type | Array Size | MKL Method Times [s] | CuPy Method Times [s] | Speed Up Factor
-- | -- | -- | -- | --
Fraunhofer | 1024 | 0.218 | 0.0261 | 8.35
Fraunhofer | 2048 | 0.755 | 0.0294 | 25.7
Fraunhofer | 4096 | 3.36 | 0.0423 | 79.4
Fresnel | 1024 | 0.714 | 0.0438 | 16.3
Fresnel | 2048 | 4.16 | 0.0845 | 49.2
Fresnel | 4096 | 17.5 | 0.225 | 77.8

</div></b>

One catch with this is none of POPPY’s display functionality is compatible with CuPy. This is because matplotlib cannot plot CuPy arrays since they are only on GPU memory. In order to obtain a Numpy array from a CuPy array, the “cupy.ndarray.get()” method can be used. So users can obtain intensity and phase arrays by adding .get().

It should be noted that the following POPPY features have not been tested for functionality with CuPy:
- PhysicalFresnelWavefront
- Active optics such as the TipTiltStage (found in active_optics.py)
- The Instrument class
- Special propagation features such as SemiAnalyticCoronagraph (found in special_prop.py)
- Sub-sampled optics such as ShackHartmannWavefrontSensor
- Floating-window centroid calculations

@douglase was also involved in the addition so I am including him in this PR. 

Let me know your thoughts and what changes need to be made.


poppy/fresnel.py
```diff
@@ -571,7 +580,8 @@ def pixelscale(self, value):
     def fov(self):
         """ FOV in arcseconds, if applicable"""
         if self.angular_coordinates:
-            return np.asarray(self.wavefront.shape) * u.pixel * self.pixelscale
+#             return np.asarray(self.wavefront.shape) * u.pixel * self.pixelscale
```
mperrin: remove unnecessary commented out line, please

poppy/accel_math.py
```diff
@@ -93,7 +102,7 @@ def _complex():
 def _r(x, y):
     """ Function to speed up computing the radius given x and y, using Numexpr if available
     Otherwise defaults to numpy. """
-    if _USE_NUMEXPR:
+    if _USE_NUMEXPR and not _USE_CUPY:
```
mperrin: I see this pattern repeated in many places (`if _USE_NUMEXPR and not _USE_CUPY:`). Are there _any_ scenarios or places in the code in which numexpr would be used in preference to cupy?   

We could consider instead adjusting the definitions of the `_USE_` variables to not have both of these set at once. e.g. 
`_USE_NUMEXPR = (conf.use_numexpr and _NUMEXPR_AVAILABLE and not _USE_CUPY)`

poppy/fresnel.py
```diff
@@ -4,13 +4,22 @@
...
 import time
 
+from importlib import reload
+
 import poppy
 from poppy.poppy_core import PlaneType, Wavefront, BaseWavefront, BaseOpticalSystem, FITSOpticalElement
 from . import utils
 from . import accel_math
+
+import numpy
```
mperrin: This is more of a question rather than a review comment, but can you explain when it's needed to use the `numpy` imported above, versus the `np` below which may be `cupy`?  I get the idea, that it lets you unambiguously access "real numpy" as opposed to the cupy version; my question is about in which cases it's necessary or desirable to do so.  Trying to improve my own understanding of how this works...

poppy/poppy_core.py
```diff
@@ -749,6 +757,8 @@ def _resample_wavefront_pixelscale(self, detector):
         Returns
         -------
         The wavefront object is modified to have the appropriate pixel scale and spatial extent.
+        
+        FIXME: Figure out a way to resample with CuPy without transferring arrays from GPU.
```
mperrin: Fixme may have been fixed already

poppy/poppy_core.py
```diff
@@ -765,42 +775,96 @@ def _resample_wavefront_pixelscale(self, detector):
...
         # Provide 2-pixel margin around image to reduce interpolation errors at edge, but also make
         # sure that image is centered properly after it gets cropped down to detector size
+        
         margin = 2
         crop_shape = [margin + shape for shape in self.wavefront.shape]
 
         # Crop wavefront down to detector size + margin- don't waste computation interpolating
         # parts of plane that get cropped out later anyways
         cropped_wf = utils.pad_or_crop_to_shape(self.wavefront, crop_shape)
+
```
mperrin: Big block of commented-out lines below here should be cleaned up

poppy/poppy_core.py
```diff
@@ -2885,6 +2961,9 @@ def __init__(self, name="unnamed optic", transmission=None, opd=None, opdunits=N
                     self.amplitude_file = transmission
                     self.amplitude, self.amplitude_header = fits.getdata(self.amplitude_file, header=True)
                     self.amplitude = self.amplitude.astype('=f8')  # ensure native byte order, see #213
+                    
+                    self.amplitude = _ncp.array(self.amplitude) # sets to CuPy array if np is cupy
+
```
mperrin: Looks like lines 2963 and 2965 both set `self.amplitude` so the first one gets overwritten and may not be needed??

poppy/poppy_core.py
```diff
@@ -2931,6 +3010,9 @@ def __init__(self, name="unnamed optic", transmission=None, opd=None, opdunits=N
                 self.opd_file = opd
                 self.opd, self.opd_header = fits.getdata(self.opd_file, header=True)
                 self.opd = self.opd.astype('=f8')
+                
+                self.opd = _ncp.array(self.opd) # sets to CuPy array if np is cupy
+
```
mperrin: Same comment as above; looks redundant.

poppy/tests/test_optics.py
```diff
@@ -10,9 +9,14 @@
...
 from .test_core import check_wavefront
 
-wavelength=1e-6
-
+from .. import accel_math
+import numpy
+if accel_math._USE_CUPY:
+    import cupy as np
+else: 
+    import numpy as np
```
mperrin: Should this switch to consistent `_ncp` usage too?

poppy/tests/test_wfe.py
```diff
@@ -325,10 +325,14 @@ def test_KolmogorovWFE_ps():
...
-        assert(np.round(ps3[0,0].value, 9) == np.round(0.0, 9))
-        assert(np.round(ps4[0,0].value, 9) == np.round(0.0, 9))
+#         assert(np.round(ps1[0,0].value, 9) == np.round(0.0, 9))
+#         assert(np.round(ps2[0,0].value, 9) == np.round(0.0, 9))
+#         assert(np.round(ps3[0,0].value, 9) == np.round(0.0, 9))
+#         assert(np.round(ps4[0,0].value, 9) == np.round(0.0, 9))
+        assert(np.round(ps1[0,0], 9) == np.round(0.0, 9))
+        assert(np.round(ps2[0,0], 9) == np.round(0.0, 9))
+        assert(np.round(ps3[0,0], 9) == np.round(0.0, 9))
+        assert(np.round(ps4[0,0], 9) == np.round(0.0, 9))
```
mperrin: what's going on here with the commented out lines?

poppy/wfe.py
```diff
@@ -191,7 +201,11 @@ class ZernikeWFE(WavefrontError):
     @utils.quantity_input(coefficients=u.meter, radius=u.meter)
     def __init__(self, name="Zernike WFE", coefficients=None, radius=None,
             aperture_stop=False, **kwargs):
-
+        
+#         accel_math.update_math_settings()
+#         global _ncp
+#         from .accel_math import _ncp
+
```
mperrin: cleanup commented lines

poppy/wfe.py
```diff
@@ -241,14 +254,26 @@ def get_opd(self, wave):
...
+                try:
+                    combined_zernikes += k_in_m * zernike.cached_zernike1(
+                        j,
+                        wave.shape,
+                        pixelscale_m,
+                        self.radius.to(u.meter).value,
+                        outside=0.0,
+                        noll_normalize=True
+                    )
+                except: # can't use cached_zernikes if the user changed between CPU and CuPy
```
mperrin: Better to have a specific exception that is caught here, like `except MathError:` or whatever the specific exception is. It's recommended Python style to not have bare `except`.

poppy/wfe.py
```diff
@@ -549,25 +576,26 @@ def get_opd(self, wave):
...
+        psd_random = _ncp.random.RandomState()
         psd_random.seed(self.seed)
-        rndm_noise = np.fft.fftshift(np.fft.fft2(psd_random.normal(size=(self.screen_size, self.screen_size))))
+        rndm_noise = _ncp.fft.fftshift(_ncp.fft.fft2(psd_random.normal(size=(self.screen_size, self.screen_size))))
         
-        psd_scaled = (np.sqrt(psd/(wave.pixelscale.value**2)) * rndm_noise)
-        opd = ((np.fft.ifft2(np.fft.ifftshift(psd_scaled)).real*surf_unit).to(u.m)).value 
+        psd_scaled = (_ncp.sqrt(psd/(wave.pixelscale.value**2)) * rndm_noise)
+#         opd = ((np.fft.ifft2(np.fft.ifftshift(psd_scaled)).real*surf_unit).to(u.m)).value 
+        opd = ((_ncp.fft.ifft2(_ncp.fft.ifftshift(psd_scaled)).real))*1e-9 # this is assuming the opd is calculated in nm
```
mperrin: Hard coded `*1e-9` here looks suspicious?

poppy/wfe.py
```diff
@@ -680,10 +708,10 @@ def get_opd(self, wave):
         
         # calculate OPD
         # Note: Factor dq consequence of delta function having a unit
-        opd_FFT = dq*a*np.sqrt(2.0*np.pi*self.dz*phi)
-        opd = npix**2*np.fft.ifft2(opd_FFT)
+        opd_FFT = dq.value*a*_ncp.sqrt(2.0*np.pi*self.dz.value*phi)
```
mperrin: Should these be `.to_value(u.m)`?

mperrin (commented):
Couple suggestions as discussed 

Swap `_ncp` to `_np` for clarity to more readers

I will investigate ways to reduce repetition of the accel_math calls, potentially via moving those into the parent class.

kian1377:
This is the current status of the tests on my machine. 
image

mperrin:
**Applause!!** Bravo that's superb to see. Thanks for slogging through this!  I will take a look here on my side as well. 

Were the fixes mostly the "usual straightforward" small details of imports or moving a couple arrays to/from the GPU? Or were there any more complex/subtle fixes needed? In either case this is great.

kian1377:
There were a few more subtle tweeks required, particularly to the ThermalBloomingWFE and to KolmogorovWFE classes. Also, some of the tests for these classes were also changed and I put some comments in the code so you can identify why I changed it. An example of this is changing the seed for the KolmogorovWFE test because the random number generator of cupy does not produce the same results as numpy, so I had to pick a seed that makes the test work or increase the number of iterations of the test. 

We can meet to discuss this in more detail whenever you are free.

mperrin:
Yes I found I had to do a similar tweak to RNG seed for the StatisticalPSDWFE test too 👍 

I hope to chat tomorrow but my availability is in flux, in part due to having a contractor coming to the house to demolish a bathroom as part of a renovation...

mperrin:
OK I have almost the same, except with the packages I have installed here it skips one fewer test and that test fails:
```
test_fresnel.py ...F...............
```

Let's see if I can figure that out.

mperrin:
I think that additional test, the fresnel PTP_short comparison test, maybe should be marked xfail (i.e. the test is expected to fail, and failures should be ignored). Even without GPU that test doesn't pass on my Mac locally now. The CI system doesn't run it since it skips if sckit-image isn't installed, and the CI environment setup doesn't include that package. I must have had that test passing at some point in the past though so it puzzles me.  

In any case that is a "low priority" test, since it tests the consistency of two different implementations of Fresnel propagation, and only one of those is generally used in poppy for most propagations. So I'm not sure if there is any practical consequences to this test not passing. On the other hand, there's got to be some reason we wrote this test...

mperrin:
Paging @douglase... Maybe you might recall something helpful about this puzzle. In testing @kian1377's CuPy GPU enhancement of poppy, I've discovered that the test `test_fresnel.test_Circular_Aperture_PTP_short()` is failing, not just on the GPU but also when I run it in plain vanilla CPU mode on my Mac. I don't know how long this has been the case that it's failing, since the CI system here on GitHub doesn't install scikit-image and that test depends on it so it gets skipped in the test runner here in Github Actions. 

Sleuthing back through git blame and past PR messages, it seems you @douglase added the initial part of what became this test in PR #196, back in 2016 (see [here](https://github.com/mperrin/poppy/pull/196/commits/bc9098ebb621c79671815c7fc4163f37216a75e8#diff-3c02055c8b6cac6450239f86250aa5638e8d79edde6ec3e45f7248352c09f6cf), though the test code was subsequent modified in various later commits.... Huh, in fact the prior most-recent modification of this test code was you @kian1377 about two years ago, in [this commit ](https://github.com/spacetelescope/poppy/pull/402/commits/4f288b626d84fef5d30aad273f6844c5f01e87e2#diff-3c02055c8b6cac6450239f86250aa5638e8d79edde6ec3e45f7248352c09f6cf) in #402. Do you recall anything about that? 

I don't think I can tell right now whether this inconsistency between the `propagate_direct` and `propagate_fresnel` functions is (a) a minor but insignificant side issue, which it would not be unreasonable to ignore, or (b) a significant test failure which is pointing us to a subtle bug in some math code, which we would be glad to have found and fixed. 

Hmmm.

kian1377:
I am working through it to see what I changed, but I dont recall ever changing anything related to that test. I remember adding a test called test_inwave_fresnel in #402, but dont remember touching any other tests. 

Also, there are a couple minor changes I also made since yesterday which I will be pushing shortly.

BradleySappington:
@kian1377 - FYI that development has been updated to settle failed tests.  Fetch/Pull required.

kian1377:
@BradleySappington , Are you talking about the issue with the scikit-image test failing? I haven't seen any commits or changes to that test so I am a little confused as to what development you are talking about.

BradleySappington:
@kian1377 I was speaking to the CI tests that have been historically failing for reasons unrelated to your branch.  Looks like you just pushed develop so some of the CI tests should now pass.

kian1377:
Ok, I saw there was a conflict so I just resolved that and updated my repo by pulling some of the other commits.

kian1377:
@mperrin , I installed scikit-image onto my desktop again and I am recreating the issue with the scikit-image test. Because this is is an issue for both CPU and GPU versions of the code, is it necessary to resolve this within this PR? I was thinking this issue should be resolved in a separate PR so there is a more clear record of its problem and the solution. 

Here is the current status of the tests on my desktop running with GPU. 
image

mperrin:
Bravo @kian1377 on getting the remainder of the tests passing! It's very nice to get the entire test suite passing again on GPU, and that gives a lot of confidence in the robustness of a PR this complicated. 

I completely concur with splitting out the failing Fresnel text (which doesn't even get ran by the Github Actions CI) to a separate task in #552. 

Let me give this one more review/reread in the next day or two (I was out last week), but I think we are very very close to pushing the merge button on this one :-)

kian1377:
Sounds good to me. I put some comments in the code so it is easy to understand why I made those changes, but feel free to delete those comments to clean up the code as you review the changes. 

One of your initial concerns that I have not addressed are those repeated lines of code that update the accel_math settings in the initialization of objects because I continued switching between GPU and CPU while I was debugging the tests, so that was a feature I kept using. 

Let me know if there are aspects of the code that I should change.

kian1377:
@mperrin One change that I just thought of that may make the code a little cleaner is instead of import cupy or numpy as _ncp, I could change it to just by xp. This way, anyone reading the code knows that xp is a stand in for np or cp. Just thought this may be a better practice to use so let me know if you want this to be changed.

mperrin:
Hi @kian1377, yes, good suggestion. I read some and found that yes using `xp` is a common/recommended idiom for "this could be numpy or cupy", and I agree it's good practice to use common conventions like that when possible.

(Remind me which code editor you're using? I remember you had asked me about PyCharm which I use. That sort of global rename refactoring is super straightforward to do in PyCharm, with automated code refactoring tools that are much more sophisticated than just a string search-and-replace. One of the many reasons I like to use PyCharm, FYI)

poppy/dms.py
```diff
@@ -284,22 +292,25 @@ def get_act_coordinates(self, one_d=False, include_transformations=False):
...
+            x_act = x_act * _ncp.ones((self.dm_shape[0], 1))
 
         if include_transformations:
             # Repeat the same transformations here as applied in AnalyticOpticalElement.get_coordinates()
             # But with opposite sense, since there the transformation is on the coordinate system and here
             # it is on the optic
             if hasattr(self, "rotation"):
-                angle = -np.deg2rad(self.rotation)
+                if isinstance(self.rotation, u.Quantity) and self.rotation.unit==u.degree:
+                    angle = -np.deg2rad(self.rotation).value
```
mperrin, line 305: Does this change have anything to do with the GPU code? Or is it an unrelated bug fix? I guess the point is that you want the variable `angle` to be a bare float rather than a Quantity in the end, yes?

In any case this could be generalized & simplified to
```
   if isinstance(self.rotation, u.Quantity):
       angle = -np.deg2rad(self.rotation.to_value(u.degree))
```
That will work for any input rotation unit.

mperrin:
@kian1377 I'm doing a full reread/review now. FYI, I assume this will be OK with you, but for minor cleanup like removing commented-out lines, I'm just going to do that myself and push to this branch; it's as easy for me to do that as it would be to flag those lines here in GitHub. 

To answer your earlier question about the repeated lines of code in the initialization of objects. I do think that's less than desirable to have blocks of repeated code. And isn't it a performance hit to have many many calls to update_math_settings which are most of the time going to be unnecessary? I don't want to get hung up on this, but I wonder if the use case of switching back and forth is rare enough or only needed in special cases for debugging, and so maybe it's not worth trying to add the extra complications to support switching arbitrarily in general?

poppy/accel_math.py
```diff
@@ -153,6 +182,8 @@ def _ifftshift(x):
         numBlocks = (int(N/blockdim[0]),int(N/blockdim[1]))
         cufftShift_2D_kernel[numBlocks, blockdim](x.ravel(),N)
         return x
+    elif _USE_CUPY:
+        return cp.fft.ifftshift(x)
     else:
         return np.fft.ifftshift(x)
```
mperrin, line 194: Couldn't we simplify this whole part to use `_ncp`, or `xp` after that switch, to call the appropriate CPU/GPU code? In other words like:
```
else:
    return xp.fft.ifftshift(x)
```

This is true for several places here in accel_math.py. 

For now I'm choosing to leave this as-is - we want to get this PR merged in rather than keep polishing indefinitely :-)

poppy/accel_math.py
```diff
@@ -255,7 +288,13 @@ def fft_2d(wavefront, forward=True, normalization=None, fftshift=True):
         event.wait()
         wavefront[:] = wf_on_gpu.get()
         del wf_on_gpu
-
+        
+    if _USE_CUPY: #########################################################################
+        do_fft = cp.fft.fft2 if forward else cp.fft.ifft2
+        if normalization is None:
+            normalization = 1./wavefront.shape[0] if forward else wavefront.shape[0]
+        wavefront = do_fft(wavefront)
```
mperrin: Another place where we could simplify - I think we don't even need an _USE_CUPY branch of the if statement. The same case could handle plain numpy and cupy using the `xp` approach.

poppy/dms.py
```diff
@@ -11,11 +11,15 @@
 
 from . import utils, accel_math, poppy_core, optics
 
+accel_math.update_math_settings()
+global _ncp, _scipy
```
mperrin: What does `global` do in this context? Normally I see `global` used within a function to allow access to a variable in the outer scope namespace. But here this is at the top of the file, outside of any function definition, so I'm not sure if it has any effect?  (I'm not saying this is wrong to have it this way, I'm asking to learn more...)

kian1377:
I just went ahead and replaced all _ncp instances with xp and fixed a couple bugs that came up after doing so. I agree that we should remove those repeated blocks of code that are for switching between packages, but I kept using them for debugging purposes so if we wait until everything else in the PR is ready to be merged, then I can delete those lines and do a final check to make sure it didn't break anything else. 

And feel free to delete some lines/comments that I had put in and push them directly. I just left those so it would be easier for you during code review.

mperrin (approved):
After lots of excellent work by @kian1377, plus several passes of close edits and review by me, I'm going to declare this good to go! A major enhancement to poppy with substantial speed improvements using GPU hardware. 

The entire test suite passes using GPU hardware on my laptop. And likewise all tests pass using CPU, locally on my laptop and also on the Github Actions CI. 


Comments below are just some minor notes and comments on places we could further tune in subsequent PRs later.

poppy/dms.py
mperrin: outdated comment from earlier in review - I took out the unnecessary global lines in one of the revisions.

mperrin:
Everything passes (including local tests on GPU hardware).

For the record the one 'failing' CI check is the code coverage, which has a slight decrease in coverage. This is not surprising since there's no way currently to test GPU code on Github Actions. (See https://github.com/orgs/github/projects/4247/views/1?filterQuery=+GPU&pane=issue&itemId=4967370; this is a "future" feature for Github Actions, with no particular timeline announced)

Good to merge!

kian1377:
Thank you for doing a thorough review! Happy to have helped and contributed
to this package once more. I can certainly address bugs or more
general code updates in future pull requests. Have a great day!

PR #509, opened 2022-06-30
Updated installation instructions

ojustino:
I removed AstroConda from the installation instructions because it only supports Python <=3.7 at this time and we are dropping 3.7 support in this release. I also added a deprecation notice to the developer instructions for new releases.


mperrin (approved):
Looks good - thanks

PR #510, opened 2022-06-30
Release 1.0.3

### LINKED ISSUE #502: factorial() won't accept floats in python >=3.10
i added python 3.10 to my test matrices and i get failures due to `poppy`. the specific error is:
```
        if _is_odd(n - m):
            return 0
        else:
            for k in range(int((n - m) / 2) + 1):
                coef = ((-1) ** k * factorial(n - k) /
>                       (factorial(k) * factorial((n + m) / 2. - k) * factorial((n - m) / 2. - k)))
E               TypeError: 'float' object cannot be interpreted as an integer

../py310-alldeps/lib/python3.10/site-packages/poppy/zernike.py:174: TypeError
```
`factorial()` won't accept floats as of python 3.10: https://docs.python.org/3/library/math.html#math.factorial.

ojustino:
Crystallizes the latest changes to the development branch.

(Via #493, would close #502.)

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -16,7 +16,7 @@ jobs:
 
           - name: Coverage test in Python 3
             os: ubuntu-latest
-            python: 3.10
+            python: '3.10'
             toxenv: py310-syn-cov
 
           - name: Check for Sphinx doc build errors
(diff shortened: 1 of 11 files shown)


mperrin (approved):
Looks good let’s do it

PR #540, opened 2023-01-31
add option to mark levels in display_ee

douglase:
current behavior when `utils.display_ee` is called is to label 50, 80, and 95% encircled energy. This breaks when a level is not encircled by the PSF. This PR adds the option for the user to specify custom levels with a keyword `levels`. if the keyword is not specified then there is no change to the existing behavior.


BradleySappington:
@douglase I just pushed changes to develop that will fix your failing tests.  If you fetch/merge we can move forward with this PR.

mperrin:
I merged the latest develop into this and have pushed it here, which should hopefully get all the tests passing.

mperrin (approved):
👍 
For the record, most of the changes in this PR as seen in GitHub are other recent PRs which I just merged into this to update it to the current develop. The only actual changes specific to this PR are those in utils.py, which I reviewed and approve.

Issue #544, opened 2023-02-08
np.float deprecated as of numpy v1.24.0


adambolton:
`dtype=np.float` as on lines 66-67 of `geometry.py` is deprecated and breaks under recent versions of numpy.

Replacement with `np.float32` or `np.float64` should solve.

I could attempt the fix here and elsewhere if given guidance as to whether to use 32 or 64.

douglase:
Since elsewhere in poppy `complex128` is needed, https://github.com/spacetelescope/poppy/blob/4e25b8ec6e886b0514215fadd215d423a1c76ae9/poppy/optics.py#L127, and the python `float` was [originally](https://docs.python.org/2/library/stdtypes.html#numeric-types-int-float-long-complex) a 64 bit double I think this would be safest to become`dtype=np.float64`.

douglase:
alternatively but with the same outcome, the default NumPy datatype is `np.float_` which is an alias to double https://numpy.org/devdocs/reference/arrays.scalars.html#numpy.float_

mperrin:
I like the suggestion of `np.float_` since that’s the default data type. 

@adambolton if you could put together a fix for this and pull request it in, that would be much appreciated! Thanks for the heads up about this change in recent numpy.

mperrin:
@BradleySappington and @obi-wan76 FYI, see this thread. This is a good example of a simple software maintenance/dependencies issue with poppy (thus also with WebbPSF) which has slipped through the cracks since Shannon’s departure. 
In fact I see there’s a dependabot PR which fails because of this numpy change, https://github.com/spacetelescope/poppy/pull/538, which could have brought this to our attention a while ago. Except nobody’s been looking at the dependabot PRs lately after Shannon left. Glad you will be able to start lending a hand for all this.

adambolton:
I'll take a run at this tomorrow (Fri 10 Feb 2023). Cheers!

BradleySappington:
Thanks for the link @mperrin, will start keeping an eye on the repository

mperrin:
Fixed via #545

Issue #535, opened 2022-11-14
Update Python versions supported 


mperrin:
Astropy will drop python 3.8 support in March 2023; we should follow along then, and drop that from the supported version and the CI. 

Conversely we should add python 3.11 to the CI now that it's out.

mperrin:
Resolved by #551

PR #545, opened 2023-02-10
replacing np.float with np.float_ per Issue 544

adambolton:
Replacing `dtype=np.float` with` dtype=np.float_` to address deprecation in Numpy as described in [Issue 544](https://github.com/spacetelescope/poppy/issues/544).

Did a search of the following type:
`grep -n np.float */*.py`
`grep -n np.float */*/*.py`

Found that this only affects `poppy/geometry.py`. It also appears in a commented-out place in `poppy/zernike.py` which I also changed jik.

### DIFF
--- a/poppy/geometry.py
+++ b/poppy/geometry.py
@@ -63,8 +63,8 @@ def _oneside(x, y0, y1, r):
     if np.isscalar(y0): y0 = np.asarray(y0)
     if np.isscalar(y1): y1 = np.asarray(y1)
     sx = x.shape
-    ans = np.zeros(sx, dtype=np.float)
-    yh = np.zeros(sx, dtype=np.float)
+    ans = np.zeros(sx, dtype=np.float_)
+    yh = np.zeros(sx, dtype=np.float_)
     to = (abs(x) >= r)
     ti = (abs(x) < r)
     if np.any(to):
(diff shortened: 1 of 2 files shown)


adambolton:
Tagging @mperrin @douglase for review. No other apparent way for me to assign. Let me know if something else is needed.

douglase (approved):
looks good to me.

BradleySappington:
@adambolton , looks like its failing from numpy.asscalar deprecation.  Docs say to [use numpy.ndarray.item() instead](https://numpy.org/doc/1.21/reference/generated/numpy.asscalar.html).

adambolton:
I think this new issue is outside the scope of the change I made, so I am not sure I can address it!

mperrin:
@BradleySappington Looks like the failure is in `astropy` rather than this package? The error traceback indicates
`astropy/units/quantity_helper/function_helpers.py`

And I see that's failing only in the CI test runner which runs the minimum versions, in this case py3.8. Which is end-of-life next month anyway, and we have an issue to drop that version from the CI system (#535). So I think it may be the case that the issue here isn't something with this PR, it's that astropy has already dropped support for py3.8, and we ought to follow suit here too.

mperrin:
I was going to wait, in the expectation that we'd merge this after updating the CI versions to get all the tests passing again. Except of course my PR to update the CI version (#550) is itself failing because of this issue. So I'm going to declare to just merge this now, ignoring the failing test on python 3.8, and then subsequently merge into #550, and hopefully that will get all this happy again.

PR #551, opened 2023-03-08
Update minimum versions across the board

BradleySappington:
Implements https://github.com/spacetelescope/poppy/issues/535. With initial changes by @mperrin 

Drops support for Python 3.8. Astropy is dropping python 3.8 support as of this month so we should follow suite.

Takes over PR and 
closes #550
closes #549 
closes #548 
closes #547 
closes #543

PR #554, opened 2023-03-24
minor: misc PEP8 code style consistency reformatting 

mperrin:
This PR reformats some code for better PEP8 style consistency. In part this is also just to test the CI tests are all working as intended post #551.


mperrin:
@BradleySappington or @obi-wan76 this is a minor / simple PR which I did in part just to test the CI tests are all running again on here (yes!). Only needs a cursory review at most, but if one of you would be so kind as to briefly review and check that off it would be most appreciated!

BradleySappington (approved):
What a delightful review. Thanks for cleaning the code up!

PR #556, opened 2023-03-25
Remove older CUDA GPU code, now that it's replaced by CuPY GPU code

mperrin:
- Remove the older/obsolete CUDA GPU code
- Add initial placeholder for GPU usage documentation


mperrin:
@kian1377 can you please see if you have a GitHub collaboration invite to this repo? I ought to be able to tag you as a reviewer for a PR after you accept that. 

This is a followup PR to merging your CuPY GPU PR. This one (a) removes the older CUDA code which is obsoleted by the new code. We had agreed to do that as a separate PR; this is that PR. And (b) I started some brief initial documentation for the CuPy code. We can leave this pretty basic & minimal for now, or choose to expand it. What do you think?  FYI, Marcio (@obi-wan76) would like to organize a release of poppy relatively soon, so my inclination is to leave the docs fairly basic for now. I assume that's all fine with you as well.

kian1377:
I just accepted the invitation. I am completely fine with leaving the docs
more basic for now, but let me know if there is something I should
contribute to the docs. As far as removing the older GPU functionality, I
think it will only involve removing some code from accel_math and
__init__.py.

marcio-melendez (changes requested):
The __init__.py file has some CUDA code, 
```
use_cuda = _config.ConfigItem(True, 'Use cuda for FFTs on GPU (assuming it' +
            'is available)?')
```
The CuPy code is there so I don't know if this could  be a conflict. I don't think this is an issue but since we are removing all CUDA code maybe we should remove it from here as well. If not necessary, I'll approve the merge as I've review all the other CUDA removal in the code and the additional placeholder for the documentation.

BradleySappington (approved):

marcio-melendez (approved):

PR #561, opened 2023-04-27
Move metadata from non-standard setup.cfg into PEP 621-compliant pyproject.toml. (546 takeover)

BradleySappington:
Move metadata from non-standard setup.cfg into PEP 621-compliant pyproject.toml
remove setup.py and setup.cfg


BradleySappington:
thanks to @KOLANICH for initial submission!

BradleySappington:
Ready for Review @ojustino @obi-wan76

KOLANICH:
BTW, `"wheel"` is not needed in `build-system.requires`, modern `setuptools` adds it itself.

pyproject.toml
```diff
@@ -1,5 +1,123 @@
...
             "wheel"]
-build-backend = 'setuptools.build_meta'
\ No newline at end of file
+build-backend = 'setuptools.build_meta'
+
+[project]
+name = "poppy"
+description = "Physical optics propagation (wavefront diffraction) for optical simulations, particularly of telescopes."
+authors = [{name = "Marshall Perrin", email = "mperrin@stsci.edu"}]
+license = {file = "LISCENCE.md"}
```
KOLANICH: IMHO it's better to have it as an SPDX ID rather than a reference to the file. BTW license files (the ones having names matching certain templates) are included into wheels automatically by `setuptools`.

pyproject.toml
```diff
@@ -1,5 +1,123 @@
 [build-system]
-requires = ["setuptools >= 38.0.0",
-            "setuptools_scm >= 2.0.0",
+requires = ["setuptools >= 61.2",
+            "setuptools_scm[toml]>=7.1",
             "wheel"]
```
KOLANICH: BTW, "wheel" is not needed in build-system.requires, modern setuptools adds it itself when it is needed (and, obviously, it is needed only when wheels are built).

BradleySappington: good catch!

docs/conf.py
```diff
@@ -13,44 +13,37 @@
...
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-import os
 import sys
-import datetime
+from datetime import datetime
+from pathlib import Path
+
 import stsci_rtd_theme
+import tomli
```
zacharyburnett:
```suggestion
if sys.version_info < (3, 11):
    import tomli as tomllib
else:
    import tomllib
```

pyproject.toml
```diff
@@ -1,5 +1,122 @@
...
+    "pytest",
+    "pytest-astropy",
+]
+docs = [
+    "nbsphinx",
+    "stsci_rtd_theme",
+    "sphinx-astropy",
+    "sphinx_rtd_theme",
+    "sphinx-automodapi",
+    "sphinx-issues",
```
zacharyburnett, line 33:
```suggestion
    "sphinx-issues",
    "tomli; python_version <\"3.11\"",
```

zacharyburnett (changes requested):
`tomli` was added (read-only) to Python 3.11 as `tomllib`

docs/conf.py
```diff
@@ -13,44 +13,37 @@
...
-# The short X.Y version.
-try:
-    version = package.__version__.split('-', 1)[0]
-    # The full version, including alpha/beta/rc tags.
-    release = package.__version__
-except AttributeError:
-    version = 'dev'
-    release = 'dev'
+with open(Path(__file__).parent.parent / "pyproject.toml", "rb") as metadata_file:
+    configuration = tomli.load(metadata_file)
```
zacharyburnett:
```suggestion
    configuration = tomllib.load(metadata_file)
```

docs/conf.py
KOLANICH: Nope.

```python
try:
  import tomllib
except ImportError:
  import tomli as tomllib
```

is better.

(if `tomllib` is manually transplanted into earlier versions, it works pretty fine)

zacharyburnett: you are correct, that covers more cases

BradleySappington:
All feedback submitted!  Thanks @KOLANICH and @zacharyburnett for the extra eyes

ojustino:
Hi @BradleySappington, I can take a look by tomorrow if @obi-wan76 doesn't get to it more quickly. It seems like the branch is in a good place at first glance.

pyproject.toml
```diff
@@ -1,5 +1,123 @@
...
+    "tomli; python_version <\"3.11\"",
+]
+
+[project.urls]
+Homepage = "https://github.com/spacetelescope/poppy"
+Documentation = "https://poppy-optics.readthedocs.io/"
+"Bug Tracker" = "https://github.com/spacetelescope/poppy/issues"
+
+[project.readme]
+text = '"POPPY (**P**\ hysical **O**\ ptics **P**\ ropagation in **Py**\ thon) is a Python package that simulates physical optical propagation including diffraction. It implements a flexible framework for modeling Fraunhofer and Fresnel diffraction and point spread function formation, particularly in the context of astronomical telescopes."'
```
ojustino: It looks like `text` here is copied from the `long_description` parameter from the former `setup.cfg`, but is there a reason why we wouldn't use `file = "README.md"` now? I think it would make things more maintainable if it works.

pyproject.toml
```diff
@@ -1,5 +1,123 @@
...
+authors = [{name = "Marshall Perrin", email = "mperrin@stsci.edu"}]
+license = {text = "BSD-3-Clause"}
+requires-python = ">=3.9"
+dependencies = [
+    "numpy>=1.20.0",
+    "scipy>=1.5.0",
+    "matplotlib>=3.2.0",
+    "astropy>=4.3.0",
+]
+dynamic = ["version"]
```
ojustino, line 18: I went through some of the example notebooks and noticed that `import poppy; poppy.__version__` gives an empty string. Working backwards, it appears this has been the case since at least v1.0.3, so you didn't introduce that here.

I do see that a `poppy/version.py` file is generated as part of the installation process, so that's working OK. Is there a simple fix to be made here that can also bring back the `__version__` attribute?

ojustino (commented):
Installation worked for Python 3.9 and 3.11. I have a couple of questions that I think are worth a look, though they shouldn't prevent a merge for too long.

pyproject.toml
zacharyburnett: that is done here:
https://github.com/spacetelescope/poppy/blob/62e94c43c657ede92b5e840abad6eaff2fe1bce1/poppy/__init__.py#L21-L24
and it looks like `poppy/version.py` looks like this:
```python
# file generated by setuptools_scm
# don't change, don't track in version control
__version__ = version = '1.0.4.dev100+ged0a8b1.d20230501'
__version_tuple__ = version_tuple = (1, 0, 4, 'dev100', 'ged0a8b1.d20230501')
```
I'm not sure what's going wrong to cause the import to fail

zacharyburnett: ah, of course; it should be `from .version import __version__`

pyproject.toml
zacharyburnett: even easier, there's a shortcut in the `readme` keyword that lets it be just a filename string instead of a table; it would be `readme = "README.md"` under the `[project]` table

pyproject.toml
BradleySappington: Good catch all, thanks for the input.  Just pushed the change

pyproject.toml
BradleySappington: changed, thanks!

ojustino (approved):
Thanks for the changes.

PR #567, opened 2023-05-16
Test readthedocs

BradleySappington:
Update configs for Readthedocs to pass builds


marcio-melendez (approved):
looks good and it was tested on the servers

PR #581, opened 2023-08-17
Fix CI and ReadtheDocs Build Failures

BradleySappington:
Various build errors from imported package updates


BradleySappington:
@obi-wan76 - some of the fixes involved changing the representation of `1.j` to just be `1j`
Please advise if this is not an acceptable fix.

BradleySappington:
This fix should also allow the dependabot builds to pass.

marcio-melendez (approved):
Looks good, I think the 1.j  to 1j should do it, it is an acceptable notation.

PR #591, opened 2023-10-05
Update utils.py - add Median Absolute Deviation option to radial_profile

saikrishanthpm:
Added ability to compute median absolute deviation (MAD) to the radial profile function. MAD is a better estimator than standard deviation for smaller datasets, being more robust to outliers.

### DIFF
--- a/poppy/tests/test_utils.py
+++ b/poppy/tests/test_utils.py
@@ -132,6 +132,15 @@ def test_radial_profile(plot=False):
     assert np.allclose(prof2, prof3)
     # TODO compare those to be near a sinc profile as expected?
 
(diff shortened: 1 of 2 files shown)


mperrin (approved):
@saikrishanthpm Thank you for this PR adding the median absolute deviation option. I've tested it and it looks good to me. 

For completeness I added to this PR some unit test code that exercises this option and the stddev option (this test could be made a lot better -- for now it just checks that the code runs without error and the output array has the correct number of bins, without any check of the correctness of the output values. But, better than no test at all!). I'll verify the CI tests all run with that addition and then merge. 

Thanks again.

PR #592, opened 2023-10-13
Fix Various issues with CI actions

BradleySappington:
Had to update minimum version of astropy
Update syntax on negative complex numbers causing freznel.py tests to fail


BradleySappington:
@braingram , this update should fix PR #585

braingram (approved):
Thanks for making these fixes. Seeing the changes I now understand why the tests were failing on #585
I tested (locally) the changes in #585 and in this PR and the `py310-numexpr-mkl-cov` tests pass. LGTM!

marcio-melendez (approved):
looks good

PR #604, opened 2023-12-20
add pie wedge segmented aperture

mperrin:
- Adds a `WedgeSegmentedCircularAperture` class, for defining pie-wedge or keystone-shaped apertures; some of the aperture geometries under consideration for HWO need this kind of aperture. 
- Also, minor enhancement to radial_profile function adding a `custom_function` parameter, which lets you compute an arbitrary function evaluated in over each radial bin, for instance the min or max at a given radius. This generalizes the prior approach used for computing std dev or median absolute deviation in each bin. \
- Tests for both the above are included. 

Example: 

Unknown-26


marcio-melendez (approved):
Looks good, working as intended, in particular, the `test_WedgeSegmentedCircularAperture `function 
image

PR #611, opened 2024-03-05
Implement WedgeSegmentedDeformableMirror

mperrin:
This implements a `WedgeSegmentedDeformableMirror`, suitable for simulating pie-wedge segmented apertures with misalignments per segment. In particular this aperture geometry is used in some exploratory cases for HWO design analyses, which this allows simulating. 

Unknown-1


BradleySappington (approved):
LGTM, thanks for the __init__ order comments

PR #617, opened 2024-04-26
misc minor fixes

mperrin:
- fix float comparison to use isclose instead of exact equality
- Allow astropy  Quantities in calc_datacube wavelengths
- delete unnecessary subdir and text file
- Update docs conf.py for consistency with other packages, and clean up some docs formatting


BradleySappington (approved):
LGTM

PR #632, opened 2024-08-12
min pin numexpr

BradleySappington:
numexpr>=2.9.0 to avoid future bugs per webbpsf issue 902:
https://github.com/spacetelescope/webbpsf/issues/902

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ authors = [
 ]
 requires-python = ">=3.10"
 dependencies = [
+    "numexpr>=2.9.0",
     "numpy>=1.20.0",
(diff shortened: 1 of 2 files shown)


mperrin (approved):
This is reminding me part of why there wasn't previously a pin for this: numexpr had been left as an *optional* dependency, not strictly required.   I think there is no longer any real need for that; it's no hurdle to install.  Just inertia that it had been left as an optional dependency only. 

It seems clearly better to have it a required dependency and be able to restrict to a known-good version, versus leave as an optional dependency and maybe an unlucky user has the buggy old version. So I'm 👍 for this.

PR #637, opened 2024-09-03
use z-range for dependencies in requirements.txt

BradleySappington:
Added ranges to allow install for z level enhancements(x.y.z) without needing push.


BradleySappington:
@mperrin @obi-wan76 - Updating requirements.txt to use range over the z level updates.  No need to do a separate push for minor things that dont affect code or are trivial bug fixes, users can install them automatically.

mperrin (approved):
Looks good to me, as discussed in Slack.

PR #639, opened 2024-09-27
Fix an issue with providing astropy Units to calc_datacube

mperrin:
This fixes an issue I found with the calc_datacube function, which currently errors if provided wavelengths as astropy Quantities: 

```
>>> import numpy as np, webbpsf, astropy.units as u

>>> nrs = webbpsf.NIRSpec()
>>> nrs.calc_datacube(wavelengths=np.array([1,2,3])*u.micron)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)

[... long trace of error text from Astropy trying to write a Quantity into a FITS header]
```

This fix makes sure the wavelength is converted to a plain float in units of meters before being saved as a header keyword. 

No change is needed in `calc_psf` which already works correctly with units. This fix is just needed for `calc_datacube` which somehow slipped through the cracks for this.

### DIFF
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -325,6 +325,10 @@ class Instrument(object):
             raise ValueError("Maximum number of wavelengths exceeded. "
                              "Cannot be more than 10,000.")
 
+        def wavelength_as_meters(wavelength):
+            """helper function to avoid trying to put a Quantity into a FITS header """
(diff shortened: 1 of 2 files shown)


marcio-melendez (approved):
Looks good, I ran the example in the PR without any error.

mperrin:
(accidentally pushed a commit to this branch that wasn't intended; I force-pushed to remove it and go back to just the intended commits. I'm going to let the CI run again, and then merge)

PR #643, opened 2024-11-22
Allow shifting a detector, in alternative to shifting a source

mperrin:
Finish implementing the long-delayed option to shift a detector, rather than shifting the point source. This is implemented as an option for offsetting a detector relative to the nominal center of a beam.  The underlying math for this has been implemented for years in the `matrixDFT.py` functions; this PR adds the small amount of code to the Detector class to pass through parameters to the matrix DFT functions. 

Conceptually, shifting a source (+dX, +dY) should be equivalent to shifting the detector (-dX, -dY). There's a unit test implemented that verifies that is the case.  (In fact, that unit test is >50% of the code in this PR)

This option may be used in certain future features that may be added to webbpsf/stpsf particularly in the context of coronagraph simulations. (We can already shift the source, and shift the coronagraph mask; this added the option to shift the detector also).

To Do: 
- [x] add astropy units handling

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -3402,15 +3402,23 @@ class Detector(OpticalElement):
 
         if offset is not None:
             if len(offset) != 2:
-                raise ValueError("If a detector offset is specified, it must be a tuple or list with 2 elements, giving the (X, Y) offsets.")
+                raise ValueError("If a detector offset is specified, it must be a tuple or list with 2 elements, "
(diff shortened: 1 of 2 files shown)


mperrin:
@BradleySappington this one is a bit more technical optics in part, but hopefully the existence of a test function that passes should suffice to verify the physics part of this works as intended.

BradleySappington (approved):
Code is clean/commented and passing tests

PR #646, opened 2024-12-10
add outfile option to calc_datacube

### LINKED ISSUE #5: outfile option doesn't work properly for the calc_datacube function
The outfile doesn't save the results from the calc_datacube function as a cube but as a PSF map. This comes from a Webbpsf helpdesk ticket by Matilde Mingozzi , INC0205772. 
In the meantime users can save the correct file with the write method outside   calc_datacube. E.g., :
```
cube = miri.calc_datacube(waves)
cube.write('filename.fits')
```

mperrin:
Add `outfile` and `overwrite` parameters to calc_datacube, consistent with their implementation in `calc_psf`. 

Fixes https://github.com/spacetelescope/stpsf/issues/5

### DIFF
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -298,7 +298,7 @@ class Instrument(object):
         else:
             return result
 
-    def calc_datacube(self, wavelengths, progressbar=False, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


marcio-melendez (approved):
Looks good

PR #654, opened 2025-01-13
Doc fixes

york-stsci:
Found the missing logo image, and added it back to static.

Also turned off class inheritance diagram generation by default.


BradleySappington (approved):
If all tests appropriately on test_readthedocs then this LGTM

PR #657, opened 2025-01-15
Updated release method notes

york-stsci:
Added a bit more information, and a few more command examples, to the poppy release procedures.


BradleySappington (approved):
Comments look great, I didn't verify formatting, but assume you did

york-stsci:
As it says. I checked that it looked generally good, but didn't notice that my single-backslashes, even in the "make it code" formatting, needed to be doubled to be visible. Fixed now.

PR #669, opened 2025-12-03
[SCSB-225] migrate `[tool.pytest.ini_options]` to native TOML

zacharyburnett:
Resolves [SCSB-225](https://jira.stsci.edu/browse/SCSB-225)

Pytest 9.0 includes a native TOML table `[tool.pytest]`, which we should use instead of `[tool.pytest.ini_options]`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,6 @@ test_outputs/
 
 # PyCharm
 .idea
+
+# misc
+uv.lock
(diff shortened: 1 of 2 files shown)


marcio-melendez:
@york-stsci should I request this via ISEB?

york-stsci:
Unless it's a really complicated pull request, you don't need to. Reviewing PRs is one of our overhead tasks in general.

york-stsci (approved):
Looks good.

PR #672, opened 2026-08-07
Code linting and formatting with Ruff

mperrin:
Code listing cleanup using Ruff. Large set of minor/trivial changes in formatting and syntax. No functional or performance changes to Poppy code. 

A ruff config with some options is added to pyproject.toml in this PR. This specifies the set of listing rules that was used to implement this. 

None of these edits are is in any way essential or high priority; this was mostly done as an exercise in learning how to use some of these tools more effectively.


marcio-melendez (approved):
This looks good and timely.

PR #673, opened 2026-08-07
import scipy.ndimage instead of the removed scipy.ndimage.interpolation

schlafly:
romanisim devdeps started failing this morning because we test against development versions of scipy, which removed the scipy.ndimage.interpolation module in preparation for 2.0.  poppy imports scipy.ndimage.interpolation but does not actually use it, so this is a simple fix.  Note that scipy.ndimage is used, so I've updated the import rather than removed it.  See https://github.com/scipy/scipy/pull/25595 .

### DIFF
--- a/poppy/dms.py
+++ b/poppy/dms.py
@@ -3,7 +3,7 @@
 
 import numpy as np
 import matplotlib.pyplot as plt
-import scipy.ndimage.interpolation
+import scipy.ndimage
 import scipy.signal
 import astropy.io.fits as fits
 import astropy.units as u
--- a/poppy/dms.py
+++ b/poppy/dms.py
@@ -7,6 +7,7 @@ from abc import ABC
 import astropy.io.fits as fits
 import astropy.units as u
 import matplotlib.pyplot as plt
+import numpy as np
 import scipy.ndimage
 import scipy.signal
 
--- a/poppy/optics.py
+++ b/poppy/optics.py
@@ -1,10 +1,3 @@
-import numpy as np
-import scipy.special
-import scipy.ndimage
-import matplotlib
-import astropy.io.fits as fits
-import astropy.units as u
-import warnings
 import logging
 import warnings
 from abc import ABC, abstractmethod
@@ -13,7 +6,7 @@ import astropy.io.fits as fits
 import astropy.units as u
 import matplotlib
 import numpy as np
-import scipy.ndimage.interpolation
+import scipy.ndimage
 import scipy.special
 
 from . import accel_math, conf, geometry, utils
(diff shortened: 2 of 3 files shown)


marcio-melendez (approved):
This looks good, thanks!

marcio-melendez:
@york-stsci can you merge this PR? Thanks!

york-stsci:
Can we please put into this thread something explaining that we understand why a lot of our CI checks are failing on this PR, and why we expect this not to be an issue once we've merged it? With that in place, I'm happy to merge.

schlafly:
FWIW, I don't understand the CI issues but they look unrelated to me.  If you'd rather I'll file an issue rather than the PR; it just looked like an easy fix.

york-stsci:
The ultimate thing here is that pkg_resources is deprecated. I can merge this, but what I would prefer (if I can get it done quickly) is to make another PR that removes pkg_resources, rebase this PR from develop after I've done that, and make sure the tests are now passing.

This seems like a simple change that shouldn't cause any errors, but right now we can't tell because the pkg_resources problem is masking any issues that might exist.

schlafly:
Because I couldn't leave well enough alone---Marshall's ruff PR also includes the fix for the pkg_resources issue, removing pkg_resources and replacing it with packaging.version.  So I think that's just an unrelated issue, but is broken in poppy at present.  pkg_resources is just old.

https://github.com/spacetelescope/poppy/pull/672/changes#diff-3c02055c8b6cac6450239f86250aa5638e8d79edde6ec3e45f7248352c09f6cfL2

I don't want to go further down this road---I'll let you decide what you want to do there?

schlafly:
> The ultimate thing here is that pkg_resources is deprecated. I can merge this, but what I would prefer (if I can get it done quickly) is to make another PR that removes pkg_resources, rebase this PR from develop after I've done that, and make sure the tests are now passing.

Sorry, replies crossed.  I agree.  There's no time crunch on my side; happy to let you take over.  Thanks!

marcio-melendez:
Sounds good, I can approve Marshall's PR that will fix the issue with the depracated function and then you can rebase and merge this PR.

york-stsci:
That sounds perfect, thank you

marcio-melendez:
Mnn different errors?

york-stsci:
Yes. Looks like we might have been masking a failure to import numpy.

york-stsci:
@schlafly looks like an easy fix adding numpy imports to two test files. You okay to do this?

schlafly:
Umm, I can, but my PR didn't touch numpy; I think that got dropped in your merge.  https://github.com/spacetelescope/poppy/pull/673/changes/dd061f70345ce36f70ddfb67b2b95075bdbcaa70

york-stsci:
I think I know what happened, and I think it was me making a mistake when I rebased this branch. Still, given that I'm guessing you already have this all checked out, can you fix it for me please?

PR #679, opened 2026-08-14
new decomposition function based on matrix multiplication from D Aros…

marcio-melendez:
…tein circa 2023
This is a new phase decomposition function as developed by David Arostein (circa 2023). The inputs are the same as `opd_expand_nonorthonormal`, so it returns the Zernike coefficients as well. 

This is another function for phase decomposition that works with orthogonal and non-orthogonal basis, so it is a general purpose function not limited by convergence on the iteration loop as in the `opd_expand_nonorthonormal` function. 

to get the coefficients:
zernike_coeffs = poppy.zernike.decompose_opd_basis_matrix(opd, nterms=nterms, aperture = aperture), with opd, nterms and aperture as usually defined. 

Using a Roman phase map, here are some examples using 
`opd_expand_nonorthonormal` --> 210 terms
image

Using the new function, `decompose_opd_basis_matrix`, same Zernike terms as before:
image

Virtually indistinguishable. However, the new function does take a longer to execute, with a faster option for the orthogonal case (see function description)

A comparison between Zernike coefficients between functions shows a bit of scatter but as shown on the comparisons above the reconstructed phase maps and residuals looks comparables. Here "old" refers to  `opd_expand_nonorthonormal` and "new" to `decompose_opd_basis_matrix`

image

### DIFF
--- a/poppy/zernike.py
+++ b/poppy/zernike.py
@@ -1093,10 +1093,6 @@ def decompose_opd_basis_matrix(opd, aperture=None, nterms=15, basis=zernike_basi
 
         which basis function to use. Defaults to Zernike
 
-    iterations : int
-
-        Number of iterations for convergence. Default is 5
-
     faster_orthogonal = bool
 
         Faster performance for orthogonal case. Default is False
--- a/poppy/zernike.py
+++ b/poppy/zernike.py
@@ -1033,73 +1033,55 @@ def decompose_opd_basis_matrix(opd, aperture=None, nterms=15, basis=zernike_basi
 
     orthonormal and *not* orthonormal. This function calculate the direct answer via matrix multiplications
 
-
-
-    This version is based on David Arostein python code on February 5, 2023. See description below.
-
-    We seek an expression like:
-
-    opd = sum(coeffs[i] * basis_set[i])
-
-    Apply the dot product with basis_set[j] to both sides:
-
-    opd . basis_set[j] = sum(coeffs[i] * basis_set[i]) . basis_set[j]
-
-    When you have an orthogonal basis, this becomes an expression for how to find coeffs[j];
-
-    this is used in the original code, in the "for" loop over iterations:
-
-    this_coeff = (opd_copy * b)[wgood].sum() / ngood
-
-    But when we don't have an orthogonal basis, the equation becomes a matrix equation for the coefficients:
-
-    B * coeffs = opd . basis_set
-
-    with:
-
-    B = a matrix with Bij = basis_set[i] . basis_set[j] = (basis_set[i] * basis_set[j])[wgood].sum()
-
-    f = opd . basis_set is a vector with elements f[i] = (opd * basis_set[i])[wgood].sum()
-
-    and you solve this system for coeffs
-
     Parameters
-
     -----------
-
     opd : 2d ndarray
-
         the OPD you want to fit
-
     aperture : 2D numpy array, optional
-
         Aperture mask for which pixels are included within the aperture.
-
         All positive nonzero values are considered within the aperture;
-
         any pixels with zero, negative, or NaN values will be considered
-
         outside the aperture, and set equal to the 'outside' parameter value.
-
         If this parameter is not set, the aperture will be inferred from
-
-        the finite (i.e. non-NaN) pixels in the OPD array.
-
+        the finite (i.e. non-NaN) pixels in the OPD array.`
     nterms : int
-
         number of terms to fit
-
     basis : function
-
         which basis function to use. Defaults to Zernike
-
     faster_orthogonal = bool
-
         Faster performance for orthogonal case. Default is False
 
+    Other Parameters
+    ----------------
+        Additional keyword arguments to this function are passed through to the `basis` callable.
+
+    Returns
+    -------
+    coeffs : list
+        List of coefficients (of length `nterms`) from which the
+        input OPD map can be constructed in the given basis.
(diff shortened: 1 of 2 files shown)


mperrin (changes requested):
@marcio-melendez thanks much for implementing this!  The algorithm and output looks good. But there are at least two things that will need to be addressed: 

1. The code appears to be double spaced for some reason? With an extra blank line, or several blank lines, between every line of code. I assume this is accidental somehow? 

2. There needs to be test code for this new functionality, in test_zernike.py. At minimum it needs to be tested that the new function works without error and returns sensible outputs. Ideally there would be some test for numerical correctness of the outputs, and/or close consistency (but not an exact match) to the outputs of the existing phase decomposition function.

marcio-melendez:
@mperrin Thanks for the comments. Let me know if this looks good to you

poppy/zernike.py
```diff
@@ -1025,6 +1025,149 @@ def decompose_opd(opd, aperture=None, nterms=15, basis=zernike_basis,
...
+    But when we don't have an orthogonal basis, the equation becomes a matrix equation for the coefficients:
+    B * coeffs = opd . basis_set
+    with:
+    B = a matrix with Bij = basis_set[i] . basis_set[j] = (basis_set[i] * basis_set[j])[wgood].sum()
+    f = opd . basis_set is a vector with elements f[i] = (opd * basis_set[i])[wgood].sum()
+    and then solve thw system for coeffs
+    """
+
+    if aperture is None:
+
```
mperrin, line 1087: The code itself still seems to be unnecessarily double spaced, with a lot of blank lines? Does it not look like that to you? 

Image

marcio-melendez: Yes, it does. Fixed. (wrong committed version)

mperrin:
This seems to work really nicely!   I did some additional testing, building on the test function you wrote. In particular I saw that test function only used a circular aperture, so it doesn't really test the case of a non-orthogonal basis. I extended the function to do that (see code below), using the `LetterFAperture` utility class to get a really-asymmetric aperture on which the Zernike polynomials are definitely far from orthonormal. 

Impressively, this new function seems to yield precise and exact retrieval of the input_coeff values even on that extreme test case. On the other hand the existing `decompose_opd_nonorthonormal_basis` does not; it only gets approximate results.   Here's my extension of your test case to test this. Take a look...  Overall I think this shows that this algorithm is indeed a significant improvement on the older iterative approach!


```
from poppy import zernike
import pytest

@pytest.mark.parametrize("aper_shape", ['circle', 'f'])
def test_decompose_opd_basis_matrix(npix = 512, nterms = 10, input_coeffs = (0.1, 0.2, 0.3, 0.4, 0.5), aper_shape='circle'):
    """
    Simple test for zernike.decompose_opd_basis_matrix
    Checks that that function can run without errors and that a
    reconstructed OPD return similar results as other functions
    """


    # Build OPD from those coefficients
    opd = zernike.compose_opd_from_basis(input_coeffs, npix=npix)

    if aper_shape=='circle':
        aperture = np.isfinite(opd).astype(float)
    elif aper_shape=='f':
        aperture = poppy.LetterFAperture().sample(npix=npix)   # Very asymmetric test aperture
        opd *= aperture
    else:
        raise ValueError('unknown aperture shape')
    opd_nan_filled = np.nan_to_num(opd)

    #Run test_decompose_opd_basis_matrix  without error
    coeffs_new = zernike.decompose_opd_basis_matrix(
        opd_nan_filled, aperture=aperture, nterms=nterms
    )

    # Check correct number of coefficients returned
    assert len(coeffs_new) == nterms

    # Recovers coefficients between test_decompose_opd_basis_matrix
    # and the actual input
    assert np.allclose(coeffs_new[0:5], input_coeffs) 
    assert np.allclose(coeffs_new[5:], 0)


    if aper_shape=='circle':
        # Some additional tests we can do on a circular aperture


        # Consistency with existing decompose_opd, which implicitly assumes orthonormal basis 
        # not necessarily identical result but similar
        coeffs_simple_algorithm = zernike.decompose_opd(opd_nan_filled, aperture=aperture, nterms=nterms)
        assert np.allclose(coeffs_new[0:5], coeffs_simple_algorithm[0:5], atol=1e-3)
    
        # Consistency with  existing decompose_opd nonorthonormal
        # not necessarily identical result but similar
        # not really "old" but other implementation
        coeffs_old = zernike.decompose_opd_nonorthonormal_basis(opd_nan_filled, aperture=aperture, nterms=nterms)
        assert np.allclose(coeffs_new[0:5], coeffs_old[0:5], atol=1e-3)

  
        # Also test the faster_orthogonal=True branch runs and gives similar results
        # when compare in the orthogonal case
        coeffs_old_orthogonal = zernike.decompose_opd(opd_nan_filled, aperture=aperture, nterms=nterms)
        coeffs_new_fast = zernike.decompose_opd_basis_matrix(
            opd_nan_filled, aperture=aperture, nterms=nterms, faster_orthogonal=True
        )
        assert np.allclose(coeffs_new_fast[0:5], coeffs_old_orthogonal[0:5], atol=1e-3)
```

marcio-melendez:
thanks for the improvement to the test. I also did not know the decorator as a parameter in pytest, very nice.

mperrin (approved):
Looks good to me now, both the code and the test.

marcio-melendez:
@york-stsci you can go ahead and merge this PR. I updated the jira legacy ticket to initiate poppy stable release. 
Thanks!

PR #680, opened 2026-08-17
Improve docstrings for poppy_core.py and optics.py

mperrin:
Docstring cleanup and consistency check.

- Fill in missing input parameters and/or missing returned outputs to doc strings
- Update parameter type hints in doc strings, particularly for the many cases in which values can be supplied as an astropy Quantity or a plan float (which gets interpreted as having some default unit)
- Implement a handful of small corrections to a few mis-statements or out-of-date statements in the doc strings.

Implemented mostly using Github Copilot, then checked, verified, and refined by me. Per ST's recent policy, I therefore also add the standard AI_POLICY.md.

### DIFF
--- a/poppy/optics.py
+++ b/poppy/optics.py
@@ -52,11 +52,11 @@ class AnalyticOpticalElement(OpticalElement):
         transmission, opd : string
             These are *not allowed* for Analytic optical elements, and this class will raise an
             error if you try to set one.
-        shift_x, shift_y : astropy Quantity or float (length or angle, depending on plane type), optional
+        shift_x, shift_y : astropy Quantity (length or angle, depending on plane type) or float, optional
             Translations of this optic. Plain floats are interpreted in meters for pupil plane
             elements or arcseconds for image plane elements; astropy Quantities will be converted
             to the appropriate unit automatically.
-        rotation : astropy Quantity or float angle, optional
+        rotation : astropy Quantity angle or float, optional
             Rotation of the optic around its center, in degrees counterclockwise, or an
             angular astropy Quantity. Note that if you apply both shift and rotation,
             the optic rotates around its own center, rather than the optical axis.
@@ -256,7 +256,7 @@ class AnalyticOpticalElement(OpticalElement):
         ----------
         what : string
             What quantity to save. See the sample function of this class
-        wavelength : astropy Quantity or float length
+        wavelength : astropy Quantity length or float
             Wavelength for evaluating the optic; passed to sample(),
             in meters (plain float) or an astropy Quantity.
         npix : integer
@@ -418,7 +418,7 @@ class ScalarOpticalPathDifference(AnalyticOpticalElement):
     ----------
     name : string, optional
         Descriptive name. Auto-generated from OPD value if not given.
-    opd : astropy Quantity or float length
+    opd : astropy Quantity length or float
         Optical path difference to apply uniformly. Plain float is interpreted as meters.
         Default is 1 micron.
     """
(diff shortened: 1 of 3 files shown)


poppy/optics.py
```diff
@@ -711,8 +771,12 @@ class RectangularFieldStop(AnalyticImagePlaneElement):
     ----------
     name : string
         Descriptive name
-    width, height: float
-        Size of the field stop, in arcseconds. Default 0.5 width, height 5.
+    width : astropy Quantity or float angle
```
york-stsci: What is a float angle?

poppy/optics.py
```diff
@@ -850,10 +914,12 @@ class AnnularFieldStop(AnalyticImagePlaneElement):
     ----------
     name : string
         Descriptive name
-    radius_inner : float
-        Radius of the central opaque region, in arcseconds. Default is 0.0 (no central opaque spot)
-    radius_outer : float
-        Radius of the circular field stop outer edge. Default is 10. Set to 0.0 for no outer edge.
+    radius_inner : astropy Quantity or float angle
```
york-stsci: Same thing. This particular description is repeated several times.

poppy/optics.py
```diff
@@ -1069,6 +1138,16 @@ class LetterFAperture(AnalyticOpticalElement):
     """ Define a capital letter F aperture. This is sometimes useful for
     disambiguating pupil orientations. See also AsymmetricParityTestAperture
     and LetterFOpticalPathDifference.
+
+    Parameters
+    ----------
+    name : string, optional
+        Descriptive name. Auto-generated if not provided.
+    radius : astropy Quantity or float length
```
york-stsci: This time "float length"

poppy/optics.py
```diff
@@ -1588,44 +1697,37 @@ def _one_aperture(self, wave, index, value=1):
...
+        Radius of the full circular aperture, in meters. Default is 1.0 m.
+    rings : int
+        Number of rings of segments. Default is 2.
+    nsections : int or list of ints
+        Number of segments per ring. If one int, the same number is used for each ring.
+        Provide a list to set different numbers per ring. Use 0 as the first element
+        to exclude the center for an on-axis aperture.
+    gap_radii : astropy Quantity length, optional
```
york-stsci, line 1714: This makes me think that the description should be "astropy quantity length or float" instead of what you get above.

mperrin:
@york-stsci yeah I struggled a bit with how best to phrase this for clarity so I appreciate the feedback. The intent was something like "either an astropy Quantity in angular units, or a floating point value which is interpreted as an angle".  Perhaps it's best to just have the type hint line strictly say "an astropy Quantity or a float", and not try to also fit in any additional information about what those numbers should represent? If you think that's clearer then I will revise

york-stsci:
I actually liked "astropy Quantity angle or float" because it tells you that if you're supplying a Quantity, it needs to be something that astropy considers an angle.

poppy/optics.py
```diff
@@ -159,13 +172,24 @@ def sample(self, wavelength=1e-6 * u.meter, npix=512, grid_size=None, what='ampl
...
-        return_scale : float
+        return_scale : bool
             if True, will return a tuple containing the desired array and a float giving the
             pixel scale.
+
+        Returns
+        -------
+        output_array : ndarray
+            The sampled optic array.
+        pixel_scale : astropy Quantity
```
york-stsci, line 191: Should there also be a unit type on this?

poppy/optics.py
```diff
@@ -232,15 +256,20 @@ def to_fits(self, outname=None, what='amplitude', wavelength=1e-6 * u.meter, npi
         ----------
         what : string
             What quantity to save. See the sample function of this class
-        wavelength : float
-            Wavelength in meters.
+        wavelength : astropy Quantity length or float
```
york-stsci, line 259: Having this description (that a float will be interpreted in meters) is really helpful. Probably we should make sure it's on all of these if we're updating the documentation.

york-stsci (approved):
WRT my second comment, I don't think I actually saw any cases where there wasn't a unit, so that's just resolved.

poppy/optics.py
mperrin: yes I tried to make sure it was indeed on all of those :-)

